PR #5, opened 2019-03-20
CSharp Loader F# Support: hello sample

avestura:
Added F# hello sample from C# sample

### DIFF
--- a/source/scripts/fsharp/hello/Hello.fs
+++ b/source/scripts/fsharp/hello/Hello.fs
@@ -1,0 +1,27 @@
+namespace Scripts
+
+type IJump =
+    abstract member Jump : unit -> int
+
+type SuperJump() =
+    interface IJump with
+        member __.Jump () = 2
+
+type TinyJump() =
+    interface IJump with
+        member __.Jump () = 1
+
+type JumpMaster() =
+    static member SuperJump () : int =
+        let sj = SuperJump() :> IJump
+        sj.Jump ()
+
+    static member TinyJump () =
+        let sj = TinyJump() :> IJump
+        sj.Jump ()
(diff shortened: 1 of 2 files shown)


viferga:
It seems not to work when trying to load it as a script. It has been tested with metacallcli:
`CS0116: A namespace cannot directly contain members such as fields or methods
CS1514: { expected
CS1022: Type or namespace definition, or end-of-file expected
CS1002: ; expected
CS1022: Type or namespace definition, or end-of-file expected
CS0116: A namespace cannot directly contain members such as fields or methods
CS1022: Type or namespace definition, or end-of-file expected
CS1002: ; expected
CS1520: Method must have a return type
CS1002: ; expected
CS1022: Type or namespace definition, or end-of-file expected
CS1514: { expected
CS1513: } expected
CS1002: ; expected
...`

This is happening because of [this method](https://github.com/metacall/core/blob/develop/source/loaders/cs_loader/netcore/source/Loader.cs#L188) only supports CSharp. An addition would be to use [F# compiler services](https://fsharp.github.io/FSharp.Compiler.Service/) to implement full support to F#. Visual Basic can be easily done too, with less work (only file extension detection and switch between [SyntaxTree](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.syntaxtree?view=roslyn-dotnet) depending on it).

Compiled DLL not tested, I will edit this post later.

0xaryan:
That's all true. Roslyn is the compiler platform for only C# and VB, thus it can't analyze F# source codes. This is a sample for testing F# support, using the C# loader for .NET Assemblies. I'll add the project file for easier compilation.

PR #6, opened 2019-03-23
Fix bootstrap.js

trgwii:


viferga:
I need some function examples to test your new features. Some functions that I can use to inspect with: optional arguments, anonymous arguments, etc.., and the expected results (names of the arguments).

I also need a small documentation to add it in the README.md so the protocol for that special cases is available for everyone.

You can use the same test functions to do the documentation.

viferga:
https://github.com/metacall/core/commit/280b2b1931bf33d760f74cef191e76bfad93fa0f

PR #11, opened 2019-05-25
Nice to have: platform independent Bash, armv6j support

grepwood:
b89932c adds support for FreeBSD by changing #!/bin/bash to #!/usr/bin/env bash
3909d46 adds armv6j support by properly handling va_list

### DIFF
--- a/configure
+++ b/configure
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # Default options
 BUILD_DIR="build"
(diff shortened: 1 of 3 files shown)


viferga:
@grepwood review if this commit works on armv6j architecture: 4292faa42a4fc3412cb7859121a195526cf6a246

viferga:
I solved it here but I forgot to close this: https://github.com/metacall/core/commit/4292faa42a4fc3412cb7859121a195526cf6a246

This approach is similar to yours but more general, so it may be able to work in multiple architectures. Wrapping the va_args unifies the access to it. I took the source from here:
https://bitbucket.org/parrastudios/argentum-online-c/src/master/common/src/Portability/VariableArguments.h

PR #22, opened 2019-08-19
Add base for NodeJS async implementation and MetaCall support for Futures.

viferga:


viferga:
At current state, the then gets executed but the resolve callback never goes back to C. It only gets executed when the thread join on the destroy method of the node loader. So this may be a race condition or deadlock caused by calling JavaScript from C without protection (i.e executing the then method without protection from an asynchronous procedure).

Documentation:
- Similar issue (executing NodeJS promise from C): https://github.com/nodejs/help/issues/1412
- Source code example (working properly): https://github.com/wilhelmmatilainen/node-addon-api-async-tsf

There is a need to test if threadsafe functions work properly. In that case, the introspection function must be changed in order to allow creating threadsafe functions on start if it detects async, because creating them while the execution may be an overhead.

Further investigation is needed to see how the callback is being passed to JS and invoked. So we should do more investigation.

PR #53, opened 2020-10-25
[WIP] Object and class support

giarve:
- [x] check why ruby test crashes on metacall_destroy(), perhaps not related to this PR (not a regression bug)
- [x] reflect_context must be refactored to accept any value (so we will be able to add classes) instead of only functions
- [x] remove py_loader_impl_validate_object unless needed for reflect_context
- [x] implement object and class metadata
- [x] fix class_object reflect test (should be easy)
- [x] maybe delete debug log statements at class introspection?
- [ ] use rb_protect() to avoid ruby crashes on API functions that throw exceptions

Commits will be squashed into a single one to clean up the PR


viferga:
Related to this:
> * [ ]  check why ruby test crashes on metacall_destroy(), perhaps not related to this PR (not a regression bug)
I found it recently but it isn't on your fork. I find how to mitigate it but I don't know the reason in detail why it happens. It seems a double free.

Here's the mitigation code in upstream:
https://github.com/metacall/core/blob/17c6613e13f68f4b55947821c528423f0148be2d/source/loaders/rb_loader/source/rb_loader_impl.c#L841

Commenting out that line probably leads to a memory leak but at least it does not generate a segmentation fault.

viferga:
There is another potential bug that will affect us when implementing multi-threading lockfree support to metacall. The implementation we are thinking about (Ludovic and me) is focused on very fast reads (calls) while having slow writes (load new functions/objects or deleting them).

This will force that functions are immutable once loaded, which is fine except from some cases, and one of them is present in Ruby Loader as a old dead code between the transition of non duck typed support to duck typed:
https://github.com/metacall/core/blob/17c6613e13f68f4b55947821c528423f0148be2d/source/loaders/rb_loader/source/rb_loader_impl.c#L330

This line is useless because we now always use values for calling convenion so we can infer the return type on the caller site (before that we were using raw pointers instead of fat pointers, so the type was impossible to be inferred when calling a duck typed function).

You can delete that line too if you want.

Issue #56, opened 2020-11-05
Improve rs_port with all of the feedback from rustaceans


viferga:
Add docs.rs and documentation. Documentation examples: https://doc.rust-lang.org/stable/rust-by-example/meta/doc.html

Feedback of @In-line and @snejugal from [t.me/rustdevs](https://t.me/rustdevs):

> call eax:
> Few suggestions. 
> 
> 1. Don't use str error  types. Use structured errors.
> 2. Any should have From/Into implementations for common types, also possibly serde serialize/deserialize
> 3. Any is confusing with standard library Any, rename it, pls.
> 4. This seems plain wrong, rust strings are not null terminated https://github.com/metacall/core/blob/7d56d8c09d58bb7b769a794ef84c48ab2b156576/source/ports/rs_port/src/lib.rs#L116
> 
> SnejUgal:
> you can write a proc macro that inspects loaded files at compile time and generates all the boilerplate
> 
> https://github.com/metacall/core/blob/7d56d8c09d58bb7b769a794ef84c48ab2b156576/source/ports/rs_port/src/lib.rs#L82-L102
> 
> all this seems to depend that these C's int, long etc are always equivalent to Rust's i32, i64 etc, while this is not true
> 
> call eax:
> Few more suggestions.
> 
> Library code can be improved further. 
> 
> 1. Use RAII pattern
> 2. Don't constrain use to specific owned types like String, use references or conversion traits AsRef, etc..
> 3. Use cargo fmt, cargo clippy
> 4. Add separate metacall-c crate for raw bindings. 
> 5. Remove unnecessary type annotations, rust can pretty much infer majority of them
> 
> For good example of idiomatic ffi you can look at neon, it for example has serde serialization for serializing type to JS Values
> 
> You use a lot of imperative code with unnecessary allocations (for loops, vectors, etc..), in Rust you can use iterators with operations like map, filter, for_each, some parts can be reduced to one liners with iterators.

lights0123:
I'd add:

should Metacall's `char` really be a Rust `char`? I think it would be better suited as a Rust `u8`, as a Rust `char` is any valid Unicode codepoint, not just a value 0-255.

In-line:
@light0123 C `char` can be two bytes too

lights0123:
[...]

PR #58, opened 2020-11-05
Rust improvements

lights0123:
Fixes some, but not all issues mentioned in #56. I made my changes in small commits, so you can see what I did step-by-step.

### DIFF
--- a/source/ports/rs_port/src/lib.rs
+++ b/source/ports/rs_port/src/lib.rs
@@ -32,7 +32,7 @@ extern "C" {
     ) -> c_int;
     fn metacall_function(cfn: *const c_char) -> *mut c_void;
     fn metacall_destroy() -> c_int;
-    fn metacallfv(func: *mut c_void, args: *mut *mut c_void) -> *mut c_void;
+    fn metacallfv_s(func: *mut c_void, args: *mut *mut c_void, size: usize) -> *mut c_void;
     fn metacall_value_create_short(s: c_short) -> *mut c_void;
     fn metacall_value_create_int(i: c_int) -> *mut c_void;
     fn metacall_value_create_long(l: c_long) -> *mut c_void;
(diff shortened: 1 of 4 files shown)


lights0123:
Note that this is a breaking change, so when you publish to crates.io, you must change the version to 0.2.0.

source/ports/rs_port/src/lib.rs
```diff
@@ -18,288 +18,303 @@
...
+                Any::Int(x) => metacall_value_create_int(*x),
+                Any::Long(x) => metacall_value_create_long(*x),
+                Any::Float(x) => metacall_value_create_float(*x),
+                Any::Double(x) => metacall_value_create_double(*x),
+                Any::Bool(x) => metacall_value_create_bool(*x as c_int),
+                Any::Char(x) => metacall_value_create_char(*x as c_char),
+                Any::Str(x) => {
+                    let st = CString::new(x.as_str()).expect("can't convert to c str");
+                    metacall_value_create_string(st.as_ptr(), x.len())
+                }
```
lights0123, line 176: The documentation isn't clear on whether `metacall_value_to_string` requires a null-terminated string, as the size is passed along with it. If it isn't required, then

```suggestion
                Any::Str(x) => metacall_value_create_string(x.as_ptr() as *const _, x.len()),
```

viferga: I have added more doc about that: https://github.com/metacall/core/commit/51b85c876df753f1b89738cc253fefb9bacd9984

I also realized that metacall_value_from_string is very dangerous if not used properly, it truncates the size if the size is smaller (without leaving null terminated string), or it leads a gap in the end if the size is bigger (but in this case it is null terminated). I don't know what implications can have this, but when using the API to implement the Ports, I try to copy values when possible and I threat it as what it is, a C API.

viferga:
> Note that this is a breaking change, so when you publish to crates.io, you must change the version to 0.2.0.

Published with version 0.2.0: https://crates.io/crates/metacall

PR #71, opened 2021-02-06
Scala port

anasbarg:


viferga:
I merged manually without noticing this PR was opened. I was testing a bug @Tabzz98 reported but apparently, it seems something related to the Docker environment because I cannot reproduce it on my computer.

Probably the shared volumes are being mounted wrongly or the volume is being cached in some way, I have no idea.

PR #79, opened 2021-03-11
feat: load tsconfig from the root of the project

flozender:
I was able to load the `tsconfig.json` file from the root of the project directory. ~~One issue that I came across was that if the `tsconfig.json` file has comments in it, it doesn't work.~~
Edit: works for all types of JSON

### Example: 

This works:
```
{
  "compilerOptions": {
    "target": "es5",
    "module": "commonjs",
    "strict": true,
    "esModuleInterop": true,
    "skipLibCheck": true,
    "forceConsistentCasingInFileNames": true,
    "outDir": "dist",
    "sourceMap": true
  }
}
```

This doesn't:
```
{
  "compilerOptions": {
    "target": "es5", /* comment */
    "module": "commonjs",
    "strict": true,
    "esModuleInterop": true,
    "skipLibCheck": true,
    "forceConsistentCasingInFileNames": true,
    "outDir": "dist",
    "sourceMap": true
  }
}
```

### DIFF
--- a/source/loaders/ts_loader/bootstrap/lib/bootstrap.ts
+++ b/source/loaders/ts_loader/bootstrap/lib/bootstrap.ts
@@ -79,8 +79,12 @@ class TypeScriptLanguageServiceHost {
 		console.log(message);
 	}
 
-	getCompilationSettings() {
-		// TODO: Make it able to load tsconfig.json files
+	getCompilationSettings(currentDirectory) {
+		if (currentDirectory) {
+			const tsconfig = path.join(currentDirectory, 'tsconfig.json');
+			return node_require(tsconfig);
+		}
+
 		const options = {
 			esModuleInterop: true,
 			jsx: 2, /* React */
@@ -167,7 +171,8 @@ function ts_loader_trampoline_initialize() {
 	registry = ts.createDocumentRegistry();
 	servicesHost = new TypeScriptLanguageServiceHost();
 	services = ts.createLanguageService(servicesHost, registry);
-	const lib = path.dirname(servicesHost.getDefaultLibFileName(servicesHost.getCompilationSettings()));
+	const currentDirectory = servicesHost.getCurrentDirectory()
+	const lib = path.dirname(servicesHost.getDefaultLibFileName(servicesHost.getCompilationSettings(currentDirectory)));
 
 	// Load TypeScript Runtime
 	fs.readdirSync(lib).map(file => {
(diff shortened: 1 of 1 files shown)


viferga:
I have a doubt about this line `'./'`:
https://github.com/metacall/core/pull/79/commits/7b2ee4301ecab210dabaaba70f3ffe7ce085a39d#diff-c3ca75029244d601e86b44b1fa2dab42a36607df0bf0d6ec295c7882c4651889R93

In the source ( https://github.com/microsoft/TypeScript/blob/5053b0b987295296108598bdfd87865d9751f237/src/compiler/commandLineParser.ts#L2360 ) we can see that the third parameter is `basePath`. To what path is this relative?

Maybe we should use the LOADER_SCRIPT_PATH environment variable like I have pointed out here: https://github.com/metacall/core/blob/cd843ef37fe2efcfb33bbb64dccf0973b52e8493/source/loaders/ts_loader/bootstrap/lib/bootstrap.ts#L94

This environment variable is the standard folder where MetaCall will load the scripts. In the CLI, it is set up to `process.cwd()` but it can be overwritten.

What do you think about it?

flozender:
> I have a doubt about this line `'./'`:
> [7b2ee43#diff-c3ca75029244d601e86b44b1fa2dab42a36607df0bf0d6ec295c7882c4651889R93](https://github.com/metacall/core/commit/7b2ee4301ecab210dabaaba70f3ffe7ce085a39d#diff-c3ca75029244d601e86b44b1fa2dab42a36607df0bf0d6ec295c7882c4651889R93)
> 
> In the source ( https://github.com/microsoft/TypeScript/blob/5053b0b987295296108598bdfd87865d9751f237/src/compiler/commandLineParser.ts#L2360 ) we can see that the third parameter is `basePath`. To what path is this relative?

Yeah, I think we should use `currentDirectory` instead of `./`

> Maybe we should use the LOADER_SCRIPT_PATH environment variable like I have pointed out here:
> 
> https://github.com/metacall/core/blob/cd843ef37fe2efcfb33bbb64dccf0973b52e8493/source/loaders/ts_loader/bootstrap/lib/bootstrap.ts#L94
> 
> This environment variable is the standard folder where MetaCall will load the scripts. In the CLI, it is set up to `process.cwd()` but it can be overwritten.
> 
> What do you think about it?

Right now I'm treating `currentDirectory` as the root of the project but it's also where I run `metacall index.ts` so I tried running it from another folder and it seemed to work fine. I don't think I know enough right now to comprehend about replacing with `LOADER_SCRIPT_PATH`.

viferga:
Fair enough, the TSX test has to be updated because the tsconfig was hardcoded and will fail. But this is another PR that I can do easily.

PR #83, opened 2021-03-14
Implement code formatting

giarve:
Last commit must run:

``` sh
cmake --build ../../build --target clang-format
```

EDIT: ready


viferga:
I have been checking the find files, you could have used this:
https://cmake.org/cmake/help/latest/command/find_program.html

It is a standard way of finding executables with CMake.

PR #91, opened 2021-03-16
Implementing Dart Loader

### LINKED ISSUE #49: Implement Dart Loader
For implementing this loader, there is good support, so probably we can do it easily. Normally loaders can be fully implemented in C, but sometimes we can implement the loader in the target language itself with a small wrapper in C. This is useful where there isn't embedding API or if you want to do it in a simpler and faster way.

I have used the mock loader as a base for the dart loader, it provides a basic implementation as an example of what it should do.

As a documentation, we are going to use this:
 - PoC: https://github.com/google/embed-dart-vm
 - Reflection / Introspection: https://api.dart.dev/stable/2.9.0/dart-mirrors/dart-mirrors-library.html

Here is the list of tasks needed to implement it.
- [ ] Implement FindDart.cmake in order to find and/or install Dart and link it properly to the project (https://github.com/metacall/core/blob/a1ada51eafd8dd0d875239439143affa7839b5c6/source/loaders/dart_loader/CMakeLists.txt#L10, https://github.com/metacall/core/blob/a1ada51eafd8dd0d875239439143affa7839b5c6/source/loaders/dart_loader/CMakeLists.txt#L145, https://github.com/metacall/core/blob/a1ada51eafd8dd0d875239439143affa7839b5c6/source/loaders/dart_loader/CMakeLists.txt#L163).
- [ ] Adapt the proof of concept to the dart loader and implement basic initialization and destruction mechanism.
- [ ] Implement a mechanism to load scripts from files.
- [ ] Implement a mechanism to introspect the loaded scripts. This can be implemented completely in Dart, but we should find a way to pass the handle loaded by the Dart VM by arguments to a Dart function and then operate with it (https://api.dart.dev/stable/2.9.0/dart-mirrors/dart-mirrors-library.html).
- [ ] Define the list of types (https://github.com/metacall/core/blob/a1ada51eafd8dd0d875239439143affa7839b5c6/source/loaders/dart_loader/source/dart_loader_impl.cc#L262).
 - [ ] Implement type conversion functions between Dart types and MetaCall types (from MetaCall values to Dart values https://github.com/metacall/core/blob/a1ada51eafd8dd0d875239439143affa7839b5c6/source/loaders/dart_loader/source/dart_loader_impl.cc#L87, and from Dart values to MetaCall values https://github.com/metacall/core/blob/a1ada51eafd8dd0d875239439143affa7839b5c6/source/loaders/dart_loader/source/dart_loader_impl.cc#L154).
 - [ ] Implement the call (it should be placed more or less here: https://github.com/metacall/core/blob/a1ada51eafd8dd0d875239439143affa7839b5c6/source/loaders/dart_loader/source/dart_loader_impl.cc#L153, the original mock loader does not have any call because it is basically a mock).
- [ ] Implement an unit test with dart loader in order to verify it works properly.
- [ ] Implement an integration test with Flutter in order to demonstrate it works. We can do this in a new repository and for example use Flutter from NodeJS. Something similar to this: https://github.com/metacall/embedding-nodejs-example.

Siddhant-K-code:
fixes #49


viferga:
Why did you close this?

Siddhant-K-code:
> Why did you close this?

mistakenly additional 2 files have added of `ts_loader` and i have del. that fork too (can't reopen ) , will add a new PR.

So, Sorry for this un-professional thing. :(

PR #109, opened 2021-03-27
LLVM Loader PoC

yashtazor:
# Description

Trying out a PoC for LLVM Loader. 

### Start (29th March)
So far, I've been able to extract the LLVM IR from the C code. 

### Update (1st April)
Currently, I've found a way to extract functions in LLVM bit code (.bc) format. We can even have this in human-readable form for interpretation purposes. If we can somehow utilize these extracted methods to extract the names, arguments and return types, PoC is almost done.

Another way is to expose the LLVM IR module in the C code itself and then pass it as parameters in various methods of LLVM API to do the extraction work for us or we can even do it manually by iterating through various blocks of the LLVM IR code. This option hasn't been explored as of now and I plan on doing it in coming days.

Finally, after extraction (introspection/reflection) is done, we can move on to working with Metacall and implementing it inside the core.

## Type of change

- [x] New Feature (PoC)

# Checklist:


viferga:
This should not be merged into the project, the updated version of the PoC can be found here: https://github.com/yashtazor/LLVM-Loader-POC

I am leaving here the zip version:
[LLVM-Loader-POC-main.zip](https://github.com/metacall/core/files/6424224/LLVM-Loader-POC-main.zip)

PR #130, opened 2021-05-04
Specified LLVM version and C++ standard

yashtazor:
# Description

Specified the LLVM version as 11.0.1 in find_package and changed the C++ standard to 17 in CMakeLists.txt. Also deleted the old FindLLVM.cmake file.

## Type of change


viferga:
Please uncomment this line whenever you can:
`//#include <llvm/Analysis/BasicAliasAnalysis.h>`

PR #148, opened 2021-06-10
Dynamic execution path and load from file function

ketangupta34:
# Description
 - java loader execution_path - Add execution path function and bootstrap_execution_path functions
 - java_load_from_file - compiling class and storing in an handle 

Fixes #(issue_no)

## Type of change

- [ ] Documentation update

# Checklist:

- [ ] My code follows the style guidelines (clang-format) of this project.
- [ ] If my change is significant or breaking, I have passed all tests with `./docker-compose.sh &> output` and attached the output.
- [ ] I have tested my code with `OPTION_BUILD_SANITIZER` and `OPTION_TEST_MEMORYCHECK`. 
- [ ] I have run `make clang-format` in order to format my code.

If you are unclear about any of the above checks, have a look at our documentation [here](https://github.com/metacall/core/blob/develop/docs/README.md#63-debugging)

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -12,3 +12,4 @@ Swarnim Arun <swarnimarun11@gmail.com>
 Ben Schattinger <developer@lights0123.com>
 Muhammad Tabaza <muhammadtabaza@gmail.com>
 Anas Al Barghouthy <anas@heavenlyx.com>
+ketangupta34 <ketangupta34@gmail.com>
--- a/source/loaders/java_loader/bootstrap/lib/bootstrap.java
+++ b/source/loaders/java_loader/bootstrap/lib/bootstrap.java
@@ -1,6 +1,10 @@
 import java.io.File;
+
 import javax.tools.*;
 import java.util.*;
+
+import java.net.URL;
+import java.net.URLClassLoader;
 
 import java.lang.reflect.Method;
 import java.lang.Object;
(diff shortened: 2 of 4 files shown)


viferga:
As we talked in Discord chat, the execution path is wrong. You are using it for storing the compilation output of the load_from_file but that's not the same as appending paths into the `classpath` which is the idea of the execution path.

The compilation path can be the same as the one that is provided by the file from load_from_file or `LOADER_SCRIPTS_PATH`.

I think we should divide this into two things, one is the path you use to build (which you need it to be in the classpath too in order to load later on the classes), and the other is the classpath itself.

This means, for example, if I have a random folder, different from `LOADER_SCRIPTS_PATH` or `LOADER_LIBRARY_PATH`, I should be able to do this:
```c
metacall_execution_path("java", "/random/folder");
metacall_load_from_package("java", "SomePrebuiltClassFromRandomFolder");
```

Something like this should be doable, not sure if we should pass a extension to the package method (like `SomePrebuiltClassFromRandomFolder.class` or `SomePrebuiltClassFromRandomFolder.jar`) but that is the general idea.

Here's an example of what I am talking about: https://stackoverflow.com/a/402570

By the way, do not focus on this too much, we can still use hardcored paths for development and review this later on once discover and calls are implemented as we talked.

viferga:
The rest is correct, you got the idea of the Handle refactor. I am going to merge it.

ketangupta34:
Okay, i got what you said, i am gonna work on the execution path as discussed and make the required changes!!

viferga:
@ketangupta34 I've got an idea while writing this, so you don't have to change a lot the path thing, let me write it down:

viferga:
Probably we can just change this:
```java
private static String executionPath;
```
By:
```java
private static String[] executionPaths;
```
Or more safely:
```java
private static Set<Path> executionPaths;
```
And once we load something, you pass the array to the URLClassLoader. What do you think about?

ketangupta34:
Yes, this approach will work better and can make the working more dynamic!!. Instead of just a single classpath we can iterate over the set and find for the pkg.

PR #149, opened 2021-06-10
WebAssembly loader: Initialization, handle loading and function discovery

Encrylize:
# Description

This PR implements initialization, handle loading (from a buffer or from a file) and function discovery.

## Type of change

- [X] New feature (non-breaking change which adds functionality)

# Checklist:

- [x] My code follows the style guidelines (clang-format) of this project.
- [ ] I have added tests/screenshots (if any) that prove my fix is effective or that my feature works. (COMING SOON)
- [x] I have run `make clang-format` in order to format my code.

### DIFF
--- a/source/loaders/wasm_loader/CMakeLists.txt
+++ b/source/loaders/wasm_loader/CMakeLists.txt
@@ -7,9 +7,7 @@ endif()
 # External dependencies
 #
 
-# TODO: https://github.com/WinBuilds/ethvm-hera/blob/master/cmake/ProjectWAVM.cmake
-
-# find_package(WAVM REQUIRED)
+# TODO: Add Wasmtime as required dependency using find_package
 
 #
 # Plugin name and options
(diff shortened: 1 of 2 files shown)


viferga:
I just realized that `wasmtime` is written in Rust. I was looking for FindWasmtime.cmake but it does not exists in any GitHub repo XD

By the way, I also found this: https://arxiv.org/pdf/2012.01032.pdf

In the benchmark section `wasmtime` seems to beat most of times `WAVM` which was my original idea, also WAVM uses a C++ API, and not the standard WASI API so it has been a good choice.

How do you think we should deal with the dependency? Apparently it has a binary installation: https://docs.wasmtime.dev/cli-install.html

We have also options like this: https://webdevolutions.blob.core.windows.net/blog/pdf/insider-series-using-rust-code-in-a-cc-project-with-cmake.pdf

But I think the latter one may imply to install all the Rust toolset which is huge, and may slow down build times in Docker or similar.

PR #240, opened 2022-01-25
[WIP] add unix function to fetch metacall custom lib path

burnerlee:


viferga:
I need you to delete all you wrote in metacallcli and squash the commits. Save the code in a backup before squashing, other contributors have lost the changes due to this.

I am writing you a test so you can verify the issue.

Also use `make clang-format` in order to lint the code.

Other recommendations, change this:
- `metacall_lib_name[17]` by `metacall_lib_name[]` and in other arrays too. It's simpler and the compiler sets the size automatically.
- The `struct metacall_lib_data` can be a static struct shared between the dynlink implementation but not exposed to the external details. You can define the string as MAX_PATH or PATH_MAX (depending on platform), so you don't need to malloc or free it.
- When you define things like:
```
#define _GNU_SOURCE
// .. some includes ...
#include <link.h>
```
It is better if you put the definition of _GNU_SOURCE just on top of the header you want to affect, in this case `link.h`, and also it's recommended to do this:
```
#ifndef _GNU_SOURCE
#define _GNU_SOURCE
#endif
#include <link.h>
```

viferga:
Here's your test: https://github.com/metacall/core/commit/082a78a0fdc309805b636d9cf80a5f00758b3906

Uncomment this for testing:
https://github.com/metacall/core/blob/082a78a0fdc309805b636d9cf80a5f00758b3906/source/tests/metacall_dynlink_path_test/source/metacall_dynlink_path_test.cpp#L38

For testing use:
make -j$(nproc) metacall-dynlink-test
ctest -VV -R metacall-dynlink-test

viferga:
@burnerlee I have done a base for you:

https://github.com/metacall/core/commit/8ad085f235b5bf23bc94c17adbace50bea267c3c

You can implement windows and macos (not sure if beos has support for it). I have tried to follow your design but abstract it in such a way that code can be reused between implementations.

You can create a new PR if you prefer it, based on my last changes and updating with your current code.

Check out also how do I generate the names of the libraries because there is a function already present for that.

viferga:
Closed: https://github.com/metacall/core/pull/244

Issue #246, opened 2022-02-14
Eval in Metacall CLI fails to parse spaces properly


AkechiShiro:
### Expected Behavior
Code with spaces should be evaluated correctly inside the Metacall CLI.

### Current Behavior
At the moment, any code containing spaces raises an error when writing directly on the command line.
```
Welcome to Tijuana, tequila, sexo & marijuana.
λ eval py print("Home World")
Error: Python Error: [Type: <class 'SyntaxError'>]: ('unterminated string literal (detected at line 1)', ('0x55ee4773ea10-0x7ffd4eb3a6c0-11-1846774869', 1, 7, 'print("Home', 1, 7))
Traceback not available
Error: Python Error: [Type: <class 'SyntaxError'>]: ('unterminated string literal (detected at line 1)', ('0x55ee4773ea10-0x7ffd4eb3a6c0-11-1846774869', 1, 7, 'print("Home', 1, 7))
Traceback not available
Script 'print("Home' eval error in loader (py)
```

### Possible Solution

Either rework the parsing of arguments or put this as a goal along in the refactoring of the Metacall CLI for GSoC 2022.

### Steps to Reproduce

1. Inside a terminal, type `metacall` or `metacallcli`.
2. Input `eval py print("Hello World")`
3. See the error.

### Context (Environment)

People could wish to evaluate some code, they're writing inside the CLI and not inside any files whatsoever.

kuraakhilesh8230:
"Hi, I am  kura akhilesh, a contributor to OpenForce 2022. I would like to work on this issue, could you please tell me how this issue works. I would be making a PR as soon as I am done with resolving the issue. Thank you"

viferga:
Solved with this: https://github.com/metacall/core/pull/482

PR #258, opened 2022-04-10
fix bug in node_loader

rxbryan:
# Description
Fixes a bug that caused `napi_create_string_utf8();`
to be called with a length value less than the size
of the character buffer.
Since the strings are `NULL` terminated, the length argument to `napi_create_string_utf8();` is replaced with `NAPI_AUTO_LENGTH`
so the size of the character buffer is automatically calculated

## Screenshot of bug
Screenshot from 2022-04-10 19-34-05

## Screenshot after fix
Screenshot from 2022-04-10 20-08-41

## Type of change

- [ ] Documentation update

# Checklist:

- [ ] If my change is significant or breaking, I have passed all tests with `./docker-compose.sh build &> output` and attached the output.
- [ ] I have tested my code with `OPTION_BUILD_SANITIZER` or `./docker-compose.sh test &> output` and `OPTION_TEST_MEMORYCHECK`.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -16,3 +16,4 @@ ketangupta34 <ketangupta34@gmail.com>
 onkardahale <dahaleonkar@gmail.com>
 Akshit Garg <garg.akshit@gmail.com>
 burnerlee <avi.aviral140@gmail.com>
+rxbryan <rxbryn@gmail.com>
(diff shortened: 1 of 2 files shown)


rxbryan:
Hi @viferga, here is a bug fix for some bug I discovered while digging through the codebase

viferga:
@rxbryan I am not sure but... may be this issue related to this?
https://github.com/metacall/core/issues/246

Maybe your solution works because it's passing the same string pointer and then your solution looks for the end char '\0'. But probably this is a bug from the parser, which is calculating wrongly the size because is tokenizing the string instead of getting it all.

PR #263, opened 2022-04-13
rust port: use macro to generate inline macros

Samyak2:
# Description

The code for each language specific proc macro (`py!{}`, `node!{}`, etc.) was being generated in a build script by substituting the name in a string (`format!()`), saving the generated code in a file and then loading it back in the code.

This was unnecessary as the code can be generated through a declarative macro (a.k.a. "macro by example").

So the following changes were made:
- Defined a new macro `gen_inline_macro` which generates an inline proc macro for the given language
- This does not use any funny string interpolation
- Called this for every language defined in the build script
- Removed build script as it was only used to generate the code

Note that the usage remains the same - we get `py!{}`, `node!{}`, etc. macros as before.

Also removed `extern crate proc_macro2;` because the `extern crate` declaration is not needed in newer Rust versions.

## Type of change

- [ ] Documentation update

This can be considered a refactor, does not fit any of the above.

# Checklist:

- [ ] If my change is significant or breaking, I have passed all tests with `./docker-compose.sh build &> output` and attached the output.
- [ ] I have tested my code with `OPTION_BUILD_SANITIZER` or `./docker-compose.sh test &> output` and `OPTION_TEST_MEMORYCHECK`.


Samyak2:
All rust port tests are passing locally now.

PR #364, opened 2023-01-06
Fix NodeJS tests in GitHub CI

pkspyder007:
# Description

Adds NodeJS tests to run on CI
This is a partial fix for now, a couple of tests are still failing related to core.

### DIFF
--- a/.github/workflows/sanitizer.yml
+++ b/.github/workflows/sanitizer.yml
@@ -22,4 +22,4 @@ jobs:
           fetch-depth: 0
 
       - name: Install, build and run sanitizer tests
-        run: ./docker-compose.sh sanitizer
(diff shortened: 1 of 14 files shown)


pkspyder007:
We are ignoring the failure of this test as of now. It requires some work in the core as suggested by @viferga 

```bash
The following tests FAILED:
Errors while running CTest
     44 - metacall-node-async-resources-test (SEGFAULT)
```

CI Logs for all other tests passing in windows for node are [here](https://github.com/pkspyder007/metacall-core/actions/runs/3860889289/jobs/6581511538)

viferga:
I want to merge this but before doing it by means of the script, it would be nice if we can implement this at CMake level, so it works out of the box.

PR #365, opened 2023-01-10
Ruby win ci tests

pkspyder007:
# Description
Adds ruby tests to win ci (bracnhed out from `node-test` which also includes nodejs test on win ci)


viferga:
I am gonna close this because it contains old changes of other commits which have been removed already, and the only thing we need is adding "ruby" in the tests.

PR #394, opened 2023-03-07
No patch ci

### LINKED ISSUE #373: Remove the patching of Find*.cmake files in Windows CI
## 🚀 Feature

https://github.com/metacall/core/blob/develop/tools/metacall-environment.ps1

This is a requirement we can do in the future, but the idea of this is to avoid overwriting those Find*.cmake files so we can also debug and maintain those CMake files. For example, right now FindNodeJS.cmake has bugs in order to find properly the node libraries once it is built.

Maybe the Python or Ruby patches can be maintained because we do not maintain them.

I am not sure if this should be the best approach but it's an idea that I have had in order to keep debugging the CMake files we maintain in the long term. We can do this at the end (once all languages are supported) because it is just an improvement.

It is also problematic in the sense that those script files have been used for developers to configure or install the project dependencies (in Linux at least for now), and if we do the patching.. we break the repository files itself, and it's a not desired behavior.

Also this approach comes from the distributable, which it makes sense because we are just packaging it and there's no need of maintaining those files, and also they are usually placed on non-standard places (usually in the same folder where everything is built), but this is not the same in the developer environment.

pkspyder007:
# Description

Removes Patching of nodejs and python in windows CI

Fixes #373


pkspyder007:
Let's keep the issue open for tracking other language support as well.

PR #396, opened 2023-03-09
TS loader dependency copying CMAKE

pkspyder007:
# Description

Splits TS loader dependency copying command into two parts like node loader


viferga:
I prefer to implement this by myself because it is a problematic feature.

PR #431, opened 2023-05-15
Wasm Windows CI

pkspyder007:
This adds wasm to windows ci along with few fixes


pkspyder007:
Yes the compiler was throwing error for non constant array size. I don't have idea about the malloc leaking I am calling free at the end.

pkspyder007:
Something I found with wasm on windows
1- Unknown array length error
2- func `i32_f32_i64_f64_ret_i32_f32_i64_f64` is retuning invalid result in test
3- scripts  are not being loaded properly, e.g. empty_module.wasm is loaded from Debug folder not from script path
4-  wasmtime dll and lib not being copied to build output

viferga:
Solved.

PR #452, opened 2023-07-05
add rpc support on win ci

pkspyder007:
Enable RPC tests on windows CI

### DIFF
--- a/.github/workflows/windows-test.yml
+++ b/.github/workflows/windows-test.yml
@@ -46,7 +46,7 @@ jobs:
       - name: Set up the environment
         run: cmd.exe /c "powershell .\tools\metacall-environment.ps1 $Env:METACALL_INSTALL_OPTIONS"
         env:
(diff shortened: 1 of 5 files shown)


viferga:
It fails on some tests, some of them are my responsibility but others are related to file loader, I have to review it carefully.

pkspyder007:
Can we remove file loader from CI and merge the changes for RPC it is passing the tests

viferga:
I have investigated a bit and I found few things:

1) It seems sys/stat.h is ported to windows, as you PRed in the file_laoder, so let's use it as it is for now, if some old version of windows breaks, we will take care eventually.
2) I have tried to improve how the dlls are installed (aka, the dependencies) rather than let the ps1 script do it, I implemented it from CMake. It can be improved by the way, I left it as TODO with links to the proper way of doing it, but it's something for now.
3) The glob test fails because glob is not implemented in windows, there was also another error with paths. I have left a TODO there too. It should work as it is now.

I am gonna close the issue because I did a lot of tests with the ci for reviewing everything and I ended up implementing it in parallel a part from this PR. But this can be closed for now, file (almost) and rpc loaders are implemented now in windows, and all tests are passing.

PR #464, opened 2023-09-16
Update Benchmark CI to use another repo

ahmedihabb2:
# Description

Now benchmark CI pushes the results to another repo to keep track of benchmark history and to keep the main repo clean

# Changelog

- [Download the previous benchmark results](https://github.com/metacall/core/blob/1aeaede8ad542bac27ebd934bbc0220fc1c9c813/.github/workflows/benchmark.yml#L69C1-L76C1) removed as we do not need cache anymore as we compare with previous results from the other repo

- [Output directory changed](https://github.com/metacall/core/blob/1aeaede8ad542bac27ebd934bbc0220fc1c9c813/.github/workflows/benchmark.yml#L92C13-L92C60) as when it was /bench/matrix.os, the benchmark action will search for previous results in /bench directory, although we store them directly on the repo using matrix.os in order to access them using metacall.github.io/benchmarks/ubuntu-latest not metacall.github.io/benchmarks/bench/ubuntu-latest 

- [Add github-token](https://github.com/metacall/core/blob/1aeaede8ad542bac27ebd934bbc0220fc1c9c813/.github/workflows/benchmark.yml#L85) to allow the action pull the specified repo

-  [ Remove skip-fetch-gh-pages](https://github.com/metacall/core/blob/1aeaede8ad542bac27ebd934bbc0220fc1c9c813/.github/workflows/benchmark.yml#L88) to allow the action fetching the specified repo and compare with the previous results

- You will need to replace any occurrence of ahmedihabb2/benchmarks with the Metacall benchmarks repo that we will store benchmarks results in it.

## Create Private Access token

- Click on your profile pic in the top right corner
- Choose settings
- Scroll down and choose developer settings from the right menu
- Choose Personal access token and click on Tokens (Classic)
- Click on Generate new token
- Specify expiration and select the repo option and this will give the token read and write access to repos
image
- Copy the token then go to the repo setting and create a secret called PUSH_TOKEN and put the copied token as a value for the secret
- Now everything should work fine

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -74,11 +74,11 @@ jobs:
             tool: 'googlecpp'
             output-file-path: ./build/benchmarks/metacall-benchmarks.json
             # Access token to deploy GitHub Pages branch
-            github-token: ${{ secrets.PUSH_TOKEN }}
+            github-token: ${{ secrets.BENCHMARKS_PUSH_TOKEN }}
             # Disable push and deploy GitHub pages branch automatically
             auto-push: false
             # Github Pages repository name
-            gh-repository: github.com/ahmedihabb2/benchmarks
+            gh-repository: github.com/metacall/core-benchmarks
             # Github Pages branch name
             gh-pages-branch: main
             # Output directory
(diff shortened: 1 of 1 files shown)


viferga:
@ahmedihabb2 it works like a charm

PR #478, opened 2024-01-24
Add a basic implementation for the Zig port with proper testings

devraymondsh:
Add a basic implementation for the Zig port with proper testing

1. Boolean support
2. Char support
3. Short support
4. Int support
5. Long support
6. Float support
7. Double support
8. String support


devraymondsh:
This is the process for creating the bindings, requiring only the Zig compiler without any external dependencies. Similar to the Rust version, it's likely unnecessary to generate them if the API remains stable. Remember to consistently use the master version of Zig.
```bash
cd source/metacall/include/metacall
zig translate-c metacall.h > metacall-bindings.zig -lc -I ../../../../build/source/metacall/include -I ../ -I ../../../../build/source/include
```

PR #489, opened 2024-02-24
Make the C loader independent from its libclang version

devraymondsh:
# Description

Make the C loader independent from its libclang version. The Ubuntu docker image no longer distributes clang-11. So we tried to upgrade it to clang-16 and realized we didn't need to change a single line of code. For that reason, we decided to make the C loader independent from the libclang version.


viferga:
This needs to be reviewed in order to enable more ubuntu versions for testing and also implement the same thing in metacall-runtime.sh.

viferga:
Version 16 segfaults, I fixed it to 14 finally.

Issue #499, opened 2024-05-13
Thread deadlock in the NodeJS loader


devraymondsh:
## 🐛 Bug Report

Thread deadlock in the NodeJS loader

#### Run the test to replicate the issue:
```bash
ctest -VV -R metacall-node-multithread-deadlock-test
```

#### Output:
```
UpdateCTestConfiguration  from :/home/raymond/projects/core/build/DartConfiguration.tcl
Parse Config file:/home/raymond/projects/core/build/DartConfiguration.tcl
UpdateCTestConfiguration  from :/home/raymond/projects/core/build/DartConfiguration.tcl
Parse Config file:/home/raymond/projects/core/build/DartConfiguration.tcl
Test project /home/raymond/projects/core/build
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 27
    Start 27: metacall-node-multithread-deadlock-test

27: Test command: /home/raymond/projects/core/build/metacall-node-multithread-deadlock-testd
27: Working Directory: /home/raymond/projects/core/build/source/tests/metacall_node_multithread_deadlock_test
27: Environment variables:
27:  LOADER_LIBRARY_PATH=/home/raymond/projects/core/build
27:  LOADER_SCRIPT_PATH=/home/raymond/projects/core/build/scripts
27:  CONFIGURATION_PATH=/home/raymond/projects/core/build/configurations/global.json
27:  SERIAL_LIBRARY_PATH=/home/raymond/projects/core/build
27:  DETOUR_LIBRARY_PATH=/home/raymond/projects/core/build
27:  PORT_LIBRARY_PATH=/home/raymond/projects/core/build
27: Test timeout computed to be: 1500
27: [==========] Running 1 test from 1 test suite.
27: [----------] Global test environment set-up.
27: [----------] 1 test from metacall_node_multithread_deadlock_test
27: [ RUN      ] metacall_node_multithread_deadlock_test.DefaultConstructor
27: Debug: MetaCall default logger to stdout initialized
27: Debug: Initializing MetaCall
27: Debug: Global configuration loaded from /home/raymond/projects/core/build/configurations/global.json
27: Debug: Loading plugin: rapid_json_seriald
27: Debug: Loading plugin symbol: dynlink_symbol_rapid_json_serial_impl_interface_singleton
27: Debug: Set MetaCall log level to Debug
[...]

PR #500, opened 2024-05-13
Replicate thread deadlock behavior in Node loader

devraymondsh:
Replicates thread deadlock behavior in Node loader. As described in #499.


viferga:
Will merge once the test passes and the patch is done.

devraymondsh:
I close this issue in favor of #502.

PR #527, opened 2024-09-27
fix(rs-port): wrong implemenetation for user data in MetacallFuture 

hulxv:
# Description

Fixes #(issue_no)

## Type of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/source/ports/rs_port/src/metacall.rs
+++ b/source/ports/rs_port/src/metacall.rs
@@ -3,7 +3,7 @@ use crate::{
     cstring_enum, parsers,
     types::{MetacallError, MetacallNull, MetacallValue},
 };
-use std::ffi::c_void;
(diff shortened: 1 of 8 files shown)


hulxv:
@devraymondsh @viferga

devraymondsh:
The X_untyped function format is designed so that the user is not forced to pass a generic and would use the value through pointer polymorphism. Adding a generic to it would not only be a noticeable breaking change but I wouldn't think it's a good idea anyway. I suggest making a new API specifically for async functions which would be something like X_async or something similar.

viferga:
I solved this with the last commits, you can check now the implementation. We should review the memory leaks now. And also maybe improve the frontend API and add HList pattern with macros for doing the call: https://github.com/metacall/core/pull/538

Issue #460, opened 2023-09-05
Importing torch crashed when metacall is imported first.


iripoll:
## 🐛 Bug Report

Importing torch crashed when metacall is imported first.

### Expected Behavior

Load torch silently.

### Current Behavior

ipython and juputer-notebook crashes when importing torch.

### Steps to Reproduce

```
$ ipython3 
Python 3.11.2 (main, Mar 13 2023, 12:18:29) [GCC 12.2.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.5.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from metacall import metacall

In [2]: import torch
Segmentation fault (core dumped)
```
Versions: 
- torch: 2.0.1+cu117
- MetaCall:   0.7.4

viferga:
@iripoll Thank you so much for your report.

This is a limitation related to metacall that cannot run outside of metacall. This problem basically happens because once it's run outside metacall, it initializes two instances of python. The solution of this has been investigated and the proposal is to detect if it's being run under an existing runtime and then relink the loaders on the fly to the process instead of loading the loader with the libpython dependency.

Here is the issue described in more detail: https://github.com/metacall/core/issues/231

The same problem happens here with NodeJS: https://github.com/metacall/core/issues/31

I have investigated this in python and it happens also due to openssl but in this case is due to different versions of crypto library which have a dependency to openssl.

Our distributable mitigates the issue but it's not ideal, eventually we will solve this completely for all runtimes. We only need a bit of magic.

I am gonna keep the issue open so we can have more reference and cases where this happens for the future.

ChinmayK0607:
This happened with me as well when importing metacall on a colab notebook.

![image](https://github.com/metacall/core/assets/114411195/f09a683d-65f1-478d-8f4f-b9d6009698e5)

viferga:
@iripoll I have some updates in this: https://github.com/metacall/core/pull/557

The PR has finally landed and it works for Linux, Windows and MacOS. I have only implemented NodeJS for now but once I implement Python, I will make a new release and I will provide an example about this, then we can check if it works.

viferga:
This is solved already for Linux and MacOS. I am going to close the issue.

Issue #231, opened 2022-01-10
Allow MetaCall to run from `python.exe` or `node.exe`


viferga:
## 🚀 Feature

This is a cool feature that I have discovered while talking to @Duckxz . This feature is also related to GSoC-22: https://github.com/metacall/gsoc-2022#polyglot-debugger

Right now MetaCall suffers from a limitation, which is, if we import metacall in node, python or whatever, it have to be run from metacall CLI itself, otherwise what happens is that metacall creates a new instance of the host language instead of reusing the one it is already running in. For example:

**duplicated.js**:
```node
const { metacall_load_from_file } = require("metacall");
metacall_load_from_file("node", ["./index.js"]); // Here metacall will launch a new instance of Node (V8 + Libuv)
```

```sh
node duplicated.js
```

To avoid this duplicated instance in the past.. we forced to use metacall to run always under metacall CLI, so the host language is already managed by metacall, and this problem won't happen. We were able to detect if we are running under metacall or not thanks to this mechanism (node):
https://github.com/metacall/core/blob/e7cd9feeb8b184acd4f1a90f0c8814a84e41be7c/source/ports/node_port/index.js#L30

When talking with @Duckxz in order to use metacall for debugging external processes (i.e attach to a running process and intercept the functions from it transparently), this problem appeared again because there's no way to do this unless we remove this limitation.

So we came up with the following ideas in order to remove this implementation.

First of all, we need a way to detect if the program is running under metacall CLI or not, this is already implemented in NodeJS as shown before. If this is detected, we should mark the specific language as "host", a flag that can be queried from the loaders at runtime, so we can act consequently depending on that. For example, if we are running in `node.exe`, we would not create a new thread with the NodeJS instance at the initialization, and just initialize `node_loader` related things as if we were already on the V8 thread, we do not need to do the trampoline magic or anything because we are already in the thread.

Another problem is the symbols. For example, `node_loader.dll` is already linked to `node.dll`, this will be highly problematic because it won't work at all. So when the host language is NodeJS (and we are running under `node.exe`), we have to use the symbols from the `node.exe` executable instead the ones pointing to `node.dll`, for this we have two options (which I don't know all implications to of all of them in detail, but I have a general idea).

1) Static, portable and safe version:
[...]

PR #533, opened 2024-11-21
Implement support for running MetaCall from `python.exe` and `node.exe`.

viferga:
This feature implements support for this: https://github.com/metacall/core/issues/231 https://github.com/metacall/core/issues/460#issuecomment-1707169103

For implementing this, there's few details we will require:
- [x] We must provide a mechanism to identify the host language
- [ ] We must define a way to initialize the loader as host
- [ ] We have to search for metacall library in the ports:
     - [x] NodeJS
     - [ ] Python
     - [ ] Ruby
     - [ ] ... any other?
- [ ] We must modify the loaders in order to support host languages:
     - [ ] NodeJS
     - [ ] Python
     - [ ] Ruby
     - [ ] ... any other?
- [ ] We have to disable linking of the dependencies of the loaders, and move them to the configuration file, then load those dependencies before we initialize the loaders. This is a trick to allow loaders to have unresolved symbols and link to the host if the host language is the one running metacall, or link them against the dependencies.
- [ ] Bootstrapping of libmetacall.so and metacall.h by using the C Loader with metacall_load_from_package, so we do not need to implement any other API if the ports, we can bootstrap all of them easily.
- [ ] Make it work for:
     - [ ] Linux
     - [ ] Windows
     - [ ] MacOS
- [ ] Add full testing including integration tests with `metacall/install` and tests with multiple versions of NodeJS (compiled statically or dynamically), Python, and with different libc (Alpine) or different glibc versions.


viferga:
I have achieved to make it work for node, there's still a lot of work to do, but the PoC verifies that it's possible:
metacall-working

viferga:
I have achieved to make it work for NodeJS, there's still a lot of work to do but this PoC demonstrates that it's possible and we will achieve it:
metacall-working

PR #536, opened 2024-11-29
added CI for MinGW

RohanKrMahato:
# Added CI for MinGW

While testing the MinGW_build_support, we needed CI to check if the implementation was correct.

## Type of change

### DIFF
--- a/.github/workflows/windows-test.yml
+++ b/.github/workflows/windows-test.yml
@@ -81,6 +81,10 @@ jobs:
           - { sys: clang64, build_type: Debug }
           - { sys: clang64, build_type: Release }
 
+    defaults:
+      run:
+        shell: msys2 {0}
+
     steps:
       - name: Check out the repository
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


viferga:
@RohanKrMahato you created the PR again but it's the same code as before, I can't see the build matrix..

viferga:
I am leaving this here for reference: https://github.com/msys2/setup-msys2/blob/main/examples/cmake.yml

RohanKrMahato:
> @RohanKrMahato you created the PR again but it's the same code as before, I can't see the build matrix..

I was testing on my local machine and different branch on the forked repo. I will push here, a little bit of work remaining :)

RohanKrMahato:
@viferga 
image
Added all the combination of Release and Debug, also added the Ctest... Sir, I have changed the Default shell from Msys2 shell to Cmd and it is now working fine. (the test in the image are red as I have tested it on a repo which don't handle MinGW cases).

RohanKrMahato:
@viferga Sir, I'm sorry for the last comment, but what I noticed that changing the default Shell can have consequences, actually what was happening ( and why i thought it was working ) was that the `cmd` for all the test in Msys2 environment was taking the same compiler.

Note: Now I changed the `Default shell to Msys2` however now `Clang64` is working fine, but not MinGW 64,32 & Ucrt64 (it is giving the following error, may be it is because of the `strict type conversion` feature of compiler).


I'm attaching the error with MinGW64, (`Ucrt` also giving the similar error)
```
types]
  134 |                 NULL, 1, 0, (LPTSTR *)lpt_str, NULL);
      |                             ^~~~~~~~~~~~~~~~~
      |                             |
      |                             CHAR ** {aka char **}
In file included from D:/a/_temp/msys64/mingw64/include/windows.h:70,
                 from D:\a\core\core\source\log\source\log_policy_stream_syslog.c:23:
D:/a/_temp/msys64/mingw64/include/winbase.h:2578:165: note: expected 'const CHAR **' {aka 'const char **'} but argument is of type 'CHAR **' {aka 'char **'}
 2578 |   WINADVAPI WINBOOL WINAPI ReportEventA (HANDLE hEventLog, WORD wType, WORD wCategory, DWORD dwEventID, PSID lpUserSid, WORD wNumStrings, DWORD dwDataSize, LPCSTR *lpStrings, LPVOID lpRawData);
      |                                                                                                                                                             ~~~~~~~~^~~~~~~~~
mingw32-make[2]: *** [source\log\CMakeFiles\log.dir\build.make:469: source/log/CMakeFiles/log.dir/source/log_policy_stream_syslog.c.obj] Error 1
mingw32-make[1]: *** [CMakeFiles\Makefile2:5501: source/log/CMakeFiles/log.dir/all] Error 2
mingw32-make: *** [Makefile:155: all] Error 2
```

PR #558, opened 2025-04-19
Fix not found metacallcli in Docker Image

Bcoderx6:


viferga:
It seems not to work, I have been trying other things but no luck yet. I am still debugging it.

PR #588, opened 2025-10-23
replace string loader tags with `Tag` enum

### LINKED ISSUE #585: rust port: use `enum` for `tag` in load module
functions like `from_single_file` in [load module](https://github.com/metacall/core/blob/057d988c43d798eb9c5494d2f372a279f4c2df9c/source/ports/rs_port/src/load.rs) use `tag` as parameter:
```rust
pub fn from_single_file(
    tag: impl ToString,
    script: impl AsRef<Path>,
) -> Result<(), MetaCallLoaderError> {
    from_file(tag, [script])
}
```
and usage of this in example:
```rust
   // Load the file
    load::from_single_file("ts", "sum.ts").unwrap();
```

the problem is, MetaCall uses a set of tags. so for example if you tried to pass `rust` it'll fail, because it is expecting `rs` as a tag parameter.

## Solution
Use `enum` for our provided tags, this will help developers predict the tag names through LSP

abd0-omar:
# Description

Use Tag enum instead of strings in `load::from_single_file`, `load::from_file` and `load::from_memory`.
Prevents runtime errors from invalid tags like "rust" instead of "rs".

Fixes #585

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update

# Checklist:

### DIFF
--- a/source/ports/rs_port/src/load.rs
+++ b/source/ports/rs_port/src/load.rs
@@ -5,28 +5,82 @@ use crate::{
 };
 use std::{
     ffi::CString,
+    fmt,
     path::{Path, PathBuf},
     ptr,
 };
 
+pub enum Tag {
+    C,
+    Cobol,
+    Crystal,
+    CSharp,
+    Dart,
+    Deno,
+    Extension,
+    File,
+    Java,
+    Julia,
+    JavaScript,
+    JSM,
+    Kind,
+    LLVM,
+    Lua,
+    Mock,
+    NodeJS,
+    Python,
+    Ruby,
+    RPC,
+    Rust,
+    TypeScript,
+    Wasm,
+}
+
+impl fmt::Display for Tag {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Tag::C => write!(f, "c"),
+            Tag::Cobol => write!(f, "cob"),
+            Tag::Crystal => write!(f, "cr"),
+            Tag::CSharp => write!(f, "cs"),
+            Tag::Dart => write!(f, "dart"),
+            Tag::Deno => write!(f, "deno"),
+            Tag::Extension => write!(f, "ext"),
+            Tag::File => write!(f, "file"),
+            Tag::Java => write!(f, "java"),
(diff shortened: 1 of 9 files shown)


source/ports/rs_port/src/load.rs
```diff
@@ -5,28 +5,82 @@ use crate::{
 };
 use std::{
     ffi::CString,
+    fmt,
     path::{Path, PathBuf},
     ptr,
 };
 
+pub enum Tag {
```
fahdfady: I think names here in the enum has to be exactly matched with the loaders names provided here https://github.com/metacall/core/tree/develop/source/loaders.

Like `Node` not `NodeJS`
`Ts` not `TypeScript`
`Rb` not `Ruby`
and provide `///` comments to show the LSP what this tag's language is, matching our Enum with the original MetaCall API. 
That's my opinion, I need to know yours.
@viferga @hulxv

source/ports/rs_port/tests/invalid_loaders_test.rs
```diff
@@ -9,25 +13,24 @@ fn invalid_loaders() {
 
     let scripts_dir = env::current_dir().unwrap().join("tests/scripts");
     let inavlid_file = scripts_dir.join("whatever.yeet");
-    let valid_file = scripts_dir.join("script.js");
```
fahdfady: Why did we change this? this is supposed to be a test that points to invalid loaders.

source/ports/rs_port/tests/invalid_loaders_test.rs
```diff
@@ -9,25 +13,24 @@ fn invalid_loaders() {
...
+    let js_file = scripts_dir.join("script.js");
 
     if let Err(MetaCallLoaderError::FileNotFound(_)) =
-        load::from_single_file("random", inavlid_file)
```
fahdfady: This too should be left to fail.
I guess the test at all will be outdated and removed because of the feature we implemented. The Enum will make us impossible to fail, compile-time.

fahdfady (changes requested):
Good Job @abd0-omar Thanks :)
I just have a few concerns to raise.

PR #600, opened 2025-12-09
Improve `metacall-sys` crate

fahdfady:
This PR aims to better `metacall-sys` dev experience. Users should not make a build script and run `metacall-sys::build` by themselves. Like other -sys crates, `metacall-sys` should link for all crates that want to link to MetaCall.

> The reason the -sys pattern exists is to ensure that if multiple crates (or multiple versions of the high level bindings) depend on the same library, you don't get two conflicting versions linked in.


josead:
@fahdfady I have done some changes, it may generate breaking changes. Now rust port works on windows. If you can solve the issue we should port it to the new changes I have done.

PR #609, opened 2026-01-02
unblock build (ReportEventA const-fix + disable plthook_detour)

Somsubhra-Nandi:
MinGW enforces strict const-correctness for Windows APIs.
ReportEventA expects LPCSTR*, but the code passed LPTSTR*,
causing a build failure in log_policy_stream_syslog.c.

This patch updates the string array type to match the Windows
API contract. The build now proceeds past the previous failure point.

### DIFF
--- a/source/detours/plthook_detour/CMakeLists.txt
+++ b/source/detours/plthook_detour/CMakeLists.txt
@@ -2,6 +2,11 @@
 if(NOT OPTION_BUILD_DETOURS OR NOT OPTION_BUILD_DETOURS_PLTHOOK)
 	return()
 endif()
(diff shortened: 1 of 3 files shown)


Somsubhra-Nandi:
This PR addresses multiple MinGW-specific build blockers:

1. Fixes strict const-correctness for ReportEventA on MinGW
   (LPCSTR vs LPTSTR mismatch in log_policy_stream_syslog.c).

2. Disables plthook_detour on MinGW and guards dependent tests,
   as it relies on Windows APIs not available under MinGW.

With these changes, the MinGW build now progresses significantly
past the previous failure point.

PR #633, opened 2026-02-21
Fix metacall-c-lib-test on macOS: MODULE library type + remove broken dlsym underscore prefix

### LINKED ISSUE #631: `metacall-c-lib-test fails`: C loader load_from_package cannot reliably find .so and .h via execution_paths
## Description

`metacall-c-lib-test` fails on a plain Debug build (no sanitizers). The test calls:

metacall_load_from_package("c", "loadtest", NULL)

and the C loader fails to locate `libloadtest.so` and `loadtest.h`.

## Root Cause

`load_from_package` in the C loader uses `execution_paths` to guess where the `.so` and `.h` are located. This mechanism was designed for script loaders where everything is co-located in one directory. For the C loader, the compiled library and the header can be in completely different paths (e.g. `lib/` vs `include/`), so the guessing fails.

## Expected Behavior

`metacall_load_from_package("c", "loadtest", NULL)` returns 0 and all test assertions pass.

## Actual Behavior

`metacall_load_from_package` returns non-zero, test fails at the first `ASSERT_EQ`.

## Steps to Reproduce
```shell
cmake -DCMAKE_BUILD_TYPE=Debug -DOPTION_BUILD_LOADERS_C=On -DOPTION_BUILD_SCRIPTS_C=On ..
make metacall-c-lib-test
ctest -VV -R metacall-c-lib-test
```
## Related

Part of the broader plan to add explicit config/flags support to `metacall_load_from_X`.

ROKUMATE:
## Summary

Fixes `metacall-c-lib-test` which was failing on macOS Debug builds. Two independent bugs were found and fixed.

Closes #631

### `source/scripts/c/loadtest/CMakeLists.txt`
- Changed `add_library` type from `SHARED` to `MODULE`  
  On macOS, CMake `SHARED` produces `.dylib` but `dynlink_impl_unix` always searches for `.so`. `MODULE` produces `.so` on all Unix platforms.
- Added `configure_file` to copy `loadtest.h` into `LOADER_SCRIPT_PATH` so the C loader can find the header alongside the library.

### `source/loaders/c_loader/source/c_loader_impl.cpp`
- Removed the `#if defined(__APPLE__)` block in `c_loader_impl_discover_signature` that prepended `_` to symbol names before calling `dlsym`.  
  On modern macOS (macOS 26+), `dlsym(handle, "_funcname")` returns `NULL`. Only `dlsym(handle, "funcname")` works. The prefix caused every function to silently fail the address lookup, leaving the module with no callable symbols.

## Additional Issues Found (not fixed in this PR)
**A. `rapid_json_serial` missing as CMake dependency**  
In a slim C-only build, `metacall-c-lib-test` fails at startup because `rapid_json_serial` is not declared as a dependency. Adding `add_dependencies(metacall-c-lib-test rapid_json_serial)` to the test's CMakeLists would fix this.

**B. `FindLibClang.cmake` sets wrong include dir**  
The finder sets `LibClang_INCLUDE_DIR` to `include/clang-c/` (the subdirectory) instead of `include/` (the parent). Since `c_loader_impl.cpp` uses `#include <clang-c/Index.h>`, the parent must be on the include path. Needs `get_filename_component(...DIRECTORY)` in the finder.

**C. Clang system headers not found during AST parse**  
`libclang` doesn't automatically find built-in headers like `stdint.h`. Any real-world header using `uint32_t`, `size_t` etc. would fail to parse correctly. The C loader's `command_line_args` should include `-isystem <clang-resource-dir>/include`.

#

## Test Result
[ PASSED ] 1 test.
1/1 Test #37: metacall-c-lib-test ... Passed 0.63 sec
100% tests passed

Tested on: macOS 26.3 (Tahoe), ARM64, Homebrew LLVM 21.1.6

### DIFF
--- a/source/loaders/c_loader/source/c_loader_impl.cpp
+++ b/source/loaders/c_loader/source/c_loader_impl.cpp
@@ -60,7 +60,7 @@ typedef struct loader_impl_c_type
 	std::vector<std::string> execution_paths;
 	std::string libtcc_runtime_path;
 
-} *loader_impl_c;
+} * loader_impl_c;
 
 struct loader_impl_c_handle_base_type;
 
@@ -73,7 +73,7 @@ typedef struct c_loader_impl_discover_visitor_data_type
 	scope sp;
 	int result;
 
-} *c_loader_impl_discover_visitor_data;
+} * c_loader_impl_discover_visitor_data;
 
 static CXChildVisitResult c_loader_impl_discover_visitor(CXCursor cursor, CXCursor, void *data);
 
@@ -90,7 +90,7 @@ public:
 
 	virtual int discover_visitor(std::vector<const char *> &command_line_args, void *data) = 0;
 
-} *loader_impl_c_handle_base;
+} * loader_impl_c_handle_base;
 
 typedef struct loader_impl_c_handle_file_type : loader_impl_c_handle_base_type
 {
@@ -171,7 +171,7 @@ private:
 		return true;
 	}
 
-} *loader_impl_c_handle_file;
+} * loader_impl_c_handle_file;
 
 typedef struct loader_impl_c_handle_memory_type : loader_impl_c_handle_base_type
 {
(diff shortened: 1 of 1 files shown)


viferga:
@ROKUMATE can you debug the last test failing in MacOS respect to the c loader?

PR #667, opened 2026-03-06
fix(rapid_json_serial): guard integer casts against silent overflow

### LINKED ISSUE #632: Bug: Silent Data Corruption When Deserializing Unsigned/Large JSON Integers
## 🐛 Bug Report

While exploring the MetaCall core codebase, I found what looks like a silent data-loss bug in the RapidJSON serial implementation worth flagging.

### Title
**Silent data corruption when deserializing unsigned or large integer values via RapidJSON**

### Expected Behavior

When MetaCall deserializes a JSON number like `3000000000` (a valid unsigned 32-bit integer), the returned value should either:
- Correctly represent the number, or
- Return `NULL` / an error if the value cannot be safely represented in the current type system

### Current Behavior

The value is silently cast to a signed `int`, giving back `-1294967296` — a completely wrong number with no error returned to the caller. Three separate cases in `source/serials/rapid_json_serial/source/rapid_json_serial_impl.cpp` all have this problem (lines 401, 410, 421), and all already carry a `TODO: Review this` comment:

```cpp
// Line 401 — unsigned int → int, no bounds check
unsigned int ui = v->GetUint();
// TODO: Review this, in case of underflow/overflow store it in a bigger type?
log_write("metacall", LOG_LEVEL_WARNING, "Casting unsigned integer to integer (posible overflow)...");
return value_create_int((int)ui);  // ← silent truncation if ui > INT_MAX

// Line 421 — uint64_t → long, no bounds check
uint64_t ui = v->GetUint64();
// TODO: Review this, in case of underflow/overflow store it in a bigger type?
return value_create_long((long)ui); // ← always wrong if ui > LONG_MAX
```

The root cause: `metacall_value.h` has **no `METACALL_UINT` type**, so the deserializer has no correct path and silently falls back to signed casts.

### Possible Solution

Add a bounds check before casting and return `NULL` (a real error) instead of a corrupted value:

```cpp
if (ui > (unsigned int)INT_MAX)
{
    log_write("metacall", LOG_LEVEL_ERROR, "Value exceeds representable range, cannot cast safely");
    return NULL;
}
return value_create_int((int)ui);
```

Longer term, adding an unsigned integer type to the MetaCall value system would be the correct fix.

### Steps to Reproduce

1. Build MetaCall with RapidJSON serial enabled (default)
2. Call `metacall_deserialize` with the JSON string `"3000000000"`
3. Read the returned value with `metacall_value_to_int()`
4. Observe: result is `-1294967296` instead of `3000000000`

### Context (Environment)

Found while exploring the core source code. This affects any cross-language call or FaaS payload where a JSON number exceeds `INT_MAX` (i.e., anything > ~2.1 billion — large IDs, millisecond timestamps, Python large ints, etc.). The silent failure is the main concern — the caller receives wrong data with no indication that anything went wrong.

### Detailed Description

Three unguarded casts exist in `rapid_json_serial_impl.cpp`:

| Line | Cast | Problem |
|---|---|---|
| 401 | `unsigned int` → `int` | Silent truncation if value > `INT_MAX` |
| 410 | `int64_t` → `long` | Silent truncation on 32-bit platforms |
| 421 | `uint64_t` → `long` | Always wrong if value > `LONG_MAX` |

All three emit only a `LOG_LEVEL_WARNING` (which most embedders will never see) and return a corrupt value instead of signalling an error.

### Possible Implementation

- **Minimal fix:** Add `if (value > MAX) { return NULL; }` guards before each cast — 3 small changes across one file
- **Proper fix:** Add `METACALL_UINT` and `METACALL_UINT64` types to the value system so unsigned numbers have a correct representation end-to-end

Happy to work on either approach. Let me know the preferred direction.

Sukuna0007Abhi:
While reading through the RapidJSON deserializer I noticed three integer branches that cast to a narrower signed type with no bounds check. All three already had a TODO: Review this comment in the code, so this is a known debt.

The concrete bug: deserializing the JSON number 3000000000 (valid uint32, but larger than INT_MAX) silently returned -1294967296 — no error, no NULL, just wrong data handed back to the caller.

What I changed (one file):

IsUint() — try int first; if the value exceeds INT_MAX, try long (captures the full uint32 range on 64-bit where long is 64-bit); if still out of range, log an error and return NULL
IsInt64() — add a runtime guard for 32-bit platforms where long < int64_t; return NULL instead of truncating
IsUint64() — return NULL when value exceeds LONG_MAX instead of wrapping to a negative number
Added <climits> for INT_MAX / LONG_MAX / LONG_MIN
Tests: added two regression cases to serial_test that reproduce the original corruption scenario and assert the corrected behaviour.

Fixes #632 

Pls review @viferga

### DIFF
--- a/source/serials/rapid_json_serial/source/rapid_json_serial_impl.cpp
+++ b/source/serials/rapid_json_serial/source/rapid_json_serial_impl.cpp
@@ -433,7 +433,6 @@ value rapid_json_serial_impl_deserialize_value(const rapidjson::Value *v)
 #if LONG_MAX < INT64_MAX
 		if (i < (int64_t)LONG_MIN || i > (int64_t)LONG_MAX)
 		{
-			log_write("metacall", LOG_LEVEL_ERROR, "64-bit signed integer value overflows long in RapidJSON implementation");
 			return (value)metacall_error_throw("RapidJSON", -1, NULL, "64-bit signed integer value overflows long in RapidJSON implementation");
 		}
 #endif
@@ -446,7 +445,6 @@ value rapid_json_serial_impl_deserialize_value(const rapidjson::Value *v)
 
 		if (ui > (uint64_t)LONG_MAX)
 		{
-			log_write("metacall", LOG_LEVEL_ERROR, "64-bit unsigned integer value overflows long in RapidJSON implementation");
 			return (value)metacall_error_throw("RapidJSON", -1, NULL, "64-bit unsigned integer value overflows long in RapidJSON implementation");
 		}
 


source/serials/rapid_json_serial/source/rapid_json_serial_impl.cpp
```diff
@@ -422,10 +444,11 @@ value rapid_json_serial_impl_deserialize_value(const rapidjson::Value *v)
 	{
 		uint64_t ui = v->GetUint64();
 
-		/* TODO: Review this, in case of underflow/overflow store it in a bigger type? */
-#if LONG_MAX < UINT64_MAX
-		log_write("metacall", LOG_LEVEL_WARNING, "Casting unsigned long to int (posible overflow) in RapidJSON implementation");
-#endif
+		if (ui > (uint64_t)LONG_MAX)
+		{
+			log_write("metacall", LOG_LEVEL_ERROR, "64-bit unsigned integer value overflows long in RapidJSON implementation");
```
viferga: you can remove the prints

PR #668, opened 2026-03-07
loader: propagate void *data through load API and expose _ex public variants

### LINKED ISSUE #634: Add optional config map to `metacall_load_from_X` API for explicit compiler/linker flags
## 🚀 Feature

Add an optional config map parameter to `metacall_load_from_X` functions so callers can pass explicit compiler flags, include paths, and library paths directly — replacing the current fragile path-guessing approach.

### Is your feature request related to a problem?

`metacall_load_from_package` and related functions have no way to accept loader-specific configuration. The C loader works around this by guessing paths from `execution_paths`, which breaks whenever headers and libraries are in different directories. There is also no way to pass compiler flags (`-std=c++17`, `-DFOO`, etc.) or extra include/library paths at call time.

This was directly exposed by #631 where `load_from_package` failed to find `libloadtest.so` and `loadtest.h` because they were in different locations than the loader expected.

### Describe the solution you'd like

Add an optional `void *config` parameter (a `metacall_value` map) to the load functions:

```c
// Current
int metacall_load_from_package(const char *tag, const char *name, void **handle);

// Proposed
int metacall_load_from_package(const char *tag, const char *name, void **handle, void *config);
```

The `config` map is loader-defined. For the C loader, supported keys would be:

| Key | Maps to |
|-----|---------|
| `"include_paths"` | `-I /path/to/include` |
| `"library_paths"` | `-L /path/to/lib` |
| `"compiler_flags"` | `-std=c++17`, `-DFOO`, etc. |

Passing `NULL` preserves the current behavior for backward compatibility.

Files to change: `metacall_loaders.h`, `metacall_loaders.c`, the loader interface in `source/loader/`, and `c_loader_impl.cpp` to consume the keys. Other loaders can add their own keys incrementally.

### Describe alternatives you've considered

- **Keep using `execution_paths` guessing** — already proven fragile (#631). Not scalable.
- **Environment variables** (`LOADER_LIBRARY_PATH`, etc.) — global and not per-call. Can't pass per-invocation flags.
- **Separate config API call before load** — adds state and ordering requirements, less clean than a single call.

### Additional context

This was discussed with @viferga as the next planned contribution after the immediate fix in #631. The config map approach is already part of the broader design direction for the `metacall_load_from_X` API.

ROKUMATE:
## Summary

Adds an optional `void *data` configuration parameter to the three core load functions
(`load_from_file`, `load_from_memory`, `load_from_package`) at every layer of the stack,
and exposes new `_ex` variants in the public C API.

Closes #634

### 1. `loader_impl_interface` — typedef update
Added `void *data` as the last parameter to the three function pointer typedefs in
`loader_impl_interface.h`, so all loader implementations share the same updated signature.

### 2. Loaders — implementation update (18 loaders)
All loader `*_impl` files updated to accept `void *data` in their
`load_from_file`, `load_from_memory`, and `load_from_package` implementations.
The parameter is currently threaded through and available for future use
(e.g. passing per-load config maps). No loader changes its existing behaviour.

Loaders updated: `c`, `cob`, `cs`, `dart`, `ext`, `file`, `java`, `jl`, `js`,
`llvm`, `lua`, `mock`, `node`, `py`, `rb`, `rpc`, `ts`, `wasm`.

### 3. `loader_impl` / `loader` — call-site propagation + `_ex` variants
Internal call sites in `loader_impl.c` and `loader.c` updated to pass `void *data`
down the chain. New `_ex` internal variants added alongside the existing functions
so callers that don't need config can continue using the original signatures unchanged.

### 4. Public API + test
Three new public functions added to `metacall.h` / `metacall.c`:

```c
int metacall_load_from_file_ex(const char *tag, const char *paths[], size_t size, void **handle, void *data);
int metacall_load_from_memory_ex(const char *tag, const char *buffer, size_t size, void **handle, void *data);
int metacall_load_from_package_ex(const char *tag, const char *path, void **handle, void *data);
```

Existing `metacall_load_from_file/memory/package` are unchanged — they call the `_ex`
variants internally with `data = NULL`.

New test: `metacall_load_ex_test` — exercises all three `_ex` functions.

## Notes

- Passing `NULL` for `data` is always valid — existing behaviour is fully preserved.
- The `void *data` value is not interpreted by any loader yet; this PR only establishes
  the plumbing. Actual config map handling can be layered on top in a follow-up.

### DIFF
--- a/source/loader/source/loader.c
+++ b/source/loader/source/loader.c
@@ -332,7 +332,7 @@ int loader_execution_path(const loader_tag tag, const loader_path path)
 	return loader_impl_execution_path(p, plugin_impl_type(p, loader_impl), path);
 }
 
-int loader_load_from_file(const loader_tag tag, const loader_path paths[], size_t size, void **handle)
+int loader_load_from_file(const loader_tag tag, const loader_path paths[], size_t size, void **handle, void *data)
 {
 	if (loader_initialize() == 1)
 	{
@@ -350,10 +350,10 @@ int loader_load_from_file(const loader_tag tag, const loader_path paths[], size_
 	/* TODO: Disable logs here until log is completely thread safe and async signal safe */
 	/* log_write("metacall", LOG_LEVEL_DEBUG, "Loading %" PRIuS " file(s) (%s) from path(s): %s ...", size, tag, paths[0]); */
 
-	return loader_impl_load_from_file(&loader_manager, p, plugin_impl_type(p, loader_impl), paths, size, handle, NULL);
+	return loader_impl_load_from_file(&loader_manager, p, plugin_impl_type(p, loader_impl), paths, size, handle, data);
 }
 
-int loader_load_from_memory(const loader_tag tag, const char *buffer, size_t size, void **handle)
+int loader_load_from_memory(const loader_tag tag, const char *buffer, size_t size, void **handle, void *data)
 {
 	if (loader_initialize() == 1)
 	{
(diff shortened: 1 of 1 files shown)


viferga:
https://github.com/metacall/core/actions/runs/22800429101/job/68279981207?pr=668

It fails to build, why you haven't even tested this on your pc or ci?

@ROKUMATE

ROKUMATE:
the Code Format check is the remaining one 
which i am not certain abt what issue its causing 
apart from that the build and other ci checks are working 

the macos checks are not running correctly because of the `.dylib` issue  which we discussed earlier

viferga:
I updated the changes on a codespace but I couldn't commit them, so I ended up creating another commit:
https://github.com/metacall/core/commit/19b907c3e175d7f81a8d18bac9b858207bbe422e

Let's see if the CI passes, docker is not working on my computer.

viferga:
It works, I am going to close the issue in favor of it. I am going to show you an example with C++ and the C loader.

ROKUMATE:
Ohkk 👍🏼

PR #677, opened 2026-03-11
fix(cli): use ASCII box characters on Windows to fix misaligned help output

### LINKED ISSUE #537: Misaligned borders when using Metacall CLI help command on Windows
## When using the help command in the MetaCall CLI on Windows, the borders of the help output are misaligned, making it difficult to read and follow. However, the borders are properly aligned and appear well-formatted on Linux and macOS.

Steps to Reproduce:

(1). Open the `MetaCall CLI` on a Windows machine, go to terminal and write `metacall` it will open the CLI in the terminal.
(2). Run the `help` command.
(3). Observe that the borders are misaligned, making the output harder to read.

Expected Behavior: The borders in the help output should be properly aligned and consistent, similar to how they appear on Linux and macOS platforms.

Terminal Screenshot on windows:
 
Screenshot 2024-12-06 075822

devs6186:
# Description

Fixes #537

> **AI assistance disclosure**: This fix was developed with AI assistance (Claude) and reviewed, tested, and submitted by @devs6186.

The `help` command in `metacallcli` outputs a bordered table using Unicode box-drawing characters. On Windows the default console code page cannot render those characters, causing misaligned and unreadable output.

## Type of change

## Problem

Every border line in the CLI help output is drawn with Unicode characters such as `┌`, `─`, `│`, `└`, `┘`. On Windows consoles running code pages 850 or 1252 these code points are either replaced with question marks or occupy a different visual width than expected, breaking the alignment of every row in the help table.

## Root Cause

The strings were hardcoded as UTF-8 byte sequences directly in the source. On Linux and macOS terminals the locale and font render them as single-width characters, so alignment holds. On Windows the default console does not interpret the UTF-8 sequences the same way, so character widths are inconsistent.

## Solution

Introduce seven `CLI_BOX_*` macros before the help function. On Windows (any of `_WIN32`, `_WIN64`, `WIN32`, `WIN64`) the macros expand to ASCII equivalents (`+`, `-`, `|`). On all other platforms they expand to the original Unicode characters. Replace every hardcoded border string in `help()` with the corresponding macro. No logic changes, only the character set used for rendering.

## Changes Made

**`source/cli/plugins/cli_core_plugin/source/cli_core_plugin.cpp`**
- Added `CLI_BOX_TL`, `CLI_BOX_TR`, `CLI_BOX_BL`, `CLI_BOX_BR`, `CLI_BOX_H`, `CLI_BOX_V`, `CLI_BOX_VR` macros with platform detection.
- Replaced all hardcoded Unicode border strings in the `help()` function with the new macros.

## Testing

On Linux: output is visually identical to before — Unicode characters still used. On Windows: borders render as `+---...---+` and `|` which aligns correctly regardless of code page setting.

## Edge Cases

- Only box-drawing characters are replaced; all content text inside the borders is unchanged.
- No behaviour change on macOS or Linux.
- The horizontal bar macro matches the original Unicode string length in visible columns so all content rows remain correctly padded.

### DIFF
--- a/source/cli/plugins/cli_core_plugin/source/cli_core_plugin.cpp
+++ b/source/cli/plugins/cli_core_plugin/source/cli_core_plugin.cpp
@@ -32,6 +32,25 @@
 #include <unordered_map>
 #include <vector>
 
+#if defined(_WIN32) || defined(_WIN64) || defined(WIN32) || defined(WIN64)
+/* Use ASCII borders on Windows so help output aligns correctly in console */
+#define CLI_BOX_TL "\t+"
+#define CLI_BOX_TR "+"
+#define CLI_BOX_BL "\t+"
+#define CLI_BOX_BR "+"
+#define CLI_BOX_H  "--------------------------------------------------------------------------------------------"
+#define CLI_BOX_V "\t|"
+#define CLI_BOX_VR "|"
+#else
+#define CLI_BOX_TL "\t┌"
+#define CLI_BOX_TR "┐"
+#define CLI_BOX_BL "\t└"
+#define CLI_BOX_BR "┘"
+#define CLI_BOX_H  "────────────────────────────────────────────────────────────────────────────────────────"
+#define CLI_BOX_V "\t│"
+#define CLI_BOX_VR "│"
+#endif
+
 /* Error messages */
 #define LOAD_ERROR		"Failed to load a script"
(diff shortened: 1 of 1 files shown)


viferga:
I added your code without the defines: https://github.com/metacall/core/commit/7ee09f1df3f0b036e8216653d25176d18fa4d30c

PR #684, opened 2026-03-13
modernize python release workflow with oidc (#14)

### LINKED ISSUE #14: Python port package publish integrated with Travis.
Reference: https://github.com/metacall/core/pull/8#issuecomment-498457578
Needs to be solved first: https://github.com/metacall/core/issues/13

avik22835:
Fixes #14 

Hi Vicente,
The python port release workflow was broken after PyPI deprecated api token auth in their update. this fixes it

changes :
- switch auth to OIDC trusted publishing (eliminates the need for stored PyPI secrets)
- trigger only on version tags, not every push. PyPI rejects duplicate version uploads so the old trigger caused noise.
- use 'python -m build' instead of the old setup.py direct call.
- fix concurrency group (previous group used pr number which is always empty on tag pushes)

One thing you'll need to do on PyPI side : add metacall/core as a trusted publisher for release-python.yml. https://docs.pypi.org/trusted-publishers/adding-a-publisher/

let me know if anything looks off

### DIFF
--- a/.github/workflows/release-python.yml
+++ b/.github/workflows/release-python.yml
@@ -1,40 +1,40 @@
 name: Release Python Package
 
 on:
-  workflow_dispatch:
-  pull_request:
+  workflow_dispatch: # Allow manual triggers for testing or emergency releases
   push:
-    branches: [ master, develop ]
-    paths:
-      - '.github/workflows/release-python.yml'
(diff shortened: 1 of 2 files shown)


viferga:
@avik22835 what do you mean? can you link here the code you're reffing to?

PR #690, opened 2026-03-15
fix(metacall): implement metacall_error_last and metacall_error_clear

### LINKED ISSUE #141: Exception & Error handling is non-existent
MetaCall currently has no mechanism to propagate errors to the caller, aside from a simple logging to notify the user. Exceptions are often part of the API, meaning that it is necessary to handle them in order to make use of these APIs properly. Therefore, being able to propagate exceptions through MetaCall is crucial for making it usable beyond trivial scenarios.

This issue tracks the error handling status in MetaCall core and is the place for discussions related to error handling. See also the [GSoC 2021 ideas page](https://github.com/metacall/gsoc-2021#implement-an-error-handling-system).

You can find some of the error handling mechanisms of language runtimes here:
* Python: https://docs.python.org/3/c-api/exceptions.html#exception-handling
* NodeJS: https://nodejs.org/api/n-api.html#n_api_error_handling
* V8: https://v8.dev/docs/embed#exceptions
* Ruby: https://silverhammermba.github.io/emberb/c/#exceptions (See #61)
* Java: https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#exceptions

.Net Core seems to have no embedding API for exception handling and will probably keep being handled via a middle-man C# code. See the [C# loader](https://github.com/metacall/core/blob/8ae04f9d851c1d732e1b03be951e2300fa77285f/source/loaders/cs_loader/netcore/source/Providers/LoaderBase.cs#L262-L284).

There are some ideas on how to go about implementing this, the straight-forward one is to introduce a new error type in the [type system](https://github.com/metacall/core/blob/develop/docs/README.md#521-type-system) with generic error information like error message and so on, catch whatever exception is thrown from the calee and throw it on the other side of the language boundaries by constructing it again. This would of course mean specialized exceptions (subclasses) won't contain additional information. See the exception_type in [reflect_exception.c](https://github.com/metacall/core/blob/0fcdbff6ca0aae5fd2d582997ddd6e1c45f59145/source/reflect/source/reflect_exception.c#L30-L37).

Another approach might be to use the class type in the type system and have proper exceptions with class hierarchies but I can't comment on it much as I have pretty limited knowledge on how the class support works in MetaCall.

We probably also want to streamline the error handling in MetaCall core itself and make sure all the necessary functions have means to propagate their error status to the caller. Note that @viferga instists on not using exceptions or setjmp/longjmp in MetaCall itself with the argument that it would break languages with coroutines.

devs6186:
# Description

`metacall_error_last()` and `metacall_error_clear()` have been stubs since they were added. `metacall_error_last()` returned 1 unconditionally; `metacall_error_clear()` was a no-op. Any caller using the errno-style error inspection pattern after a failed `metacall()` call got nothing.

This implements both functions and adds `metacall_error_set_last()` as the entry point called by the core when a loader returns a TYPE_THROWABLE value.

Fixes #141

## Root Cause

The thread-local last-error slot was never implemented. `metacall_error_from_value()` exists and works correctly for callers that inspect the return value directly, but `metacall_error_last()` is the documented API for callers that want to check the error without inspecting the return value — the same pattern as `errno` / `strerror`. Without the slot, any C caller using the documented API got no information.

Additionally, `metacallfv_s` (the core function-call dispatch path reached by `metacallv`, `metacallv_s`, `metacallfv`, `metacallfv_s`) never checked whether `function_call` returned a TYPE_THROWABLE to populate the slot.

## Solution

**`metacall_error.c`:**
- Add a `static PORTABILITY_THREAD_LOCAL value metacall_error_last_v` slot using the existing cross-platform TLS macro (`__thread` on GCC/Clang, `__declspec(thread)` on MSVC)
- Implement `metacall_error_last()`: checks if the slot is non-NULL, delegates to `metacall_error_from_value()`
- Implement `metacall_error_clear()`: destroys the stored value and nulls the slot
- Implement `metacall_error_set_last()`: validates the input is TYPE_THROWABLE, destroys any existing stored value, copies the new value with `value_type_copy()` (which increments the exception reference count so the copy is independent of the caller's copy)

**`metacall_error.h`:**
- Add `metacall_error_set_last(void *v)` declaration with doc comment

**`metacall.c` (`metacallfv_s`):**
- After `function_call` returns, check `type_id_throwable(value_type_id(ret)) == 0` and call `metacall_error_set_last(ret)`. The copy stored in the slot is independent; the caller still owns the original return value.

## Changes Made

- `source/metacall/include/metacall/metacall_error.h` — add `metacall_error_set_last` declaration
- `source/metacall/source/metacall_error.c` — implement all three functions, add `PORTABILITY_THREAD_LOCAL` slot
- `source/metacall/source/metacall.c` — TYPE_THROWABLE detection in `metacallfv_s`

## Testing

The `metacall_python_exception_test` already tests `metacall_error_from_value()` which exercises the same exception extraction path that `metacall_error_last()` delegates to. The new functions can be verified by calling `metacall()` on a function that throws, then calling `metacall_error_last(&ex)` without inspecting the return value — it should now return 0 and populate `ex` with the exception data.

## Edge Cases

- `metacall_error_set_last(NULL)` is a no-op (guarded at entry)
- `metacall_error_set_last(non-throwable)` is a no-op (type check at entry)
- Calling `metacall_error_last` before any exception has occurred returns 1 (slot is NULL)
- `value_type_copy` on a TYPE_THROWABLE creates a new throwable that increments the inner exception's reference count — destroying either copy does not invalidate the other
- Only `metacallfv_s` sets the slot in this PR; the variadic call paths (`metacall()`, `metacallt()`, `metacallht_s()`) do not set it yet and can be wired in follow-up work

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

# Checklist

- [x] Self-review performed
- [x] Code commented in hard-to-understand areas
- [x] No new warnings
- [x] Tests: existing exception tests validate the underlying extraction path; the new API can be tested by calling metacall_error_last() after a throwing function call

> Assisted with CLAUDE.md

### DIFF
--- a/source/metacall/include/metacall/metacall_error.h
+++ b/source/metacall/include/metacall/metacall_error.h
@@ -106,6 +106,18 @@ METACALL_API int metacall_error_last(metacall_exception ex);
 */
 METACALL_API void metacall_error_clear(void);
 
+/**
+*  @brief
+*    Record a throwable value as the last error for the calling thread. Called
+*    internally by metacallfv_s when a loader returns a TYPE_THROWABLE value.
+*    May also be called by ports or loaders that want to set the last error
+*    directly. Copies @v so the caller retains ownership of the original value.
+*
+*  @param[in] v
+*    Throwable value to record; must be of type TYPE_THROWABLE
+*/
+METACALL_API void metacall_error_set_last(void *v);
+
 #ifdef __cplusplus
 }
 #endif
--- a/source/metacall/source/metacall.c
+++ b/source/metacall/source/metacall.c
@@ -1144,6 +1144,11 @@ void *metacallfv_s(void *func, void *args[], size_t size)
 
 		ret = function_call(f, args, size);
 
+		if (ret != NULL && type_id_throwable(value_type_id(ret)) == 0)
+		{
+			metacall_error_set_last(ret);
+		}
+
 		if (ret != NULL)
 		{
 			type t = signature_get_return(s);
--- a/source/metacall/source/metacall_error.c
+++ b/source/metacall/source/metacall_error.c
@@ -24,7 +24,13 @@
 
 #include <reflect/reflect_value_type.h>
 
+#include <portability/portability_compiler.h>
+
 #include <log/log.h>
+
+/* -- Private Variables -- */
+
+static PORTABILITY_THREAD_LOCAL value metacall_error_last_v = NULL;
 
 /* -- Methods -- */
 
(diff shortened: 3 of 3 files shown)


viferga:
I have to review this in detail, I'm putting it as draft meanwhile.

PR #692, opened 2026-03-15
reflect: implement subtyping API for typed containers

aditya-8787:
This PR resolves an existing TODO in 

source/reflect/include/reflect/reflect_type.h
 that had been commented out since the reflect module was created(2024).

Changes
reflect_type.h

Added #include <adt/adt_vector.h>
Removed TODO comment
Declared type_subtype(type t) and type_subtype_push(type t, type subtype)

reflect_type.c

Added vector subtypes to struct type_type
Lazy init in type_subtype, NULL-safe cleanup in type_destroy

Notes

Fully backwards compatible no existing signatures changed
No loader changes in this PR populating subtypes per language runtime is left for follow-up PRs
All existing reflect tests pass

### DIFF
--- a/source/reflect/source/reflect_type.c
+++ b/source/reflect/source/reflect_type.c
@@ -122,7 +122,7 @@ type_impl type_derived(type t)
 	return NULL;
 }
 
-vector type_subtype(type t)
+vector type_subtypes(type t)
 {
 	if (t != NULL)
 	{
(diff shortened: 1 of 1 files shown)


viferga:
This is interesting but I am not sure if we should define it as a new type called template or this kind of design is better.

@trgwii what do you think about this?

I can merge it anyway but the code is not tested, add tests @aditya-8787

PR #704, opened 2026-03-21
ci: add Valgrind memcheck CI support and fix memory leaks

MrSpideyNihal:
## Summary
Adds Valgrind memcheck CI pipeline following the same pattern as address-sanitizer, and fixes memory leaks found by Valgrind.

### CI — Valgrind memcheck
- docker-compose.sh: added sub_test_memcheck() following sub_coverage() pattern
- metacall-environment.sh: added sub_memcheck() to install valgrind on Debian/Ubuntu
- metacall-configure.sh: added memcheck option with -DOPTION_TEST_MEMORYCHECK=On
- linux-test.yml: added valgrind-memcheck job calling ./docker-compose.sh test-memcheck

### Fix — detour_unload() memory leak
- Added free(handle) in source/detour/source/detour.c
- Fixes detour-test: 0 defects (was 1 IPW)
- Fixes metacall-fork-test: 0 defects (was 2 memory leaks)

### Fix — backtrace plugin suppression
- Added Valgrind suppression file for metacall-backtrace-plugin-test
- Reduces 1121 potential leaks to 0

### DIFF
--- a/tools/metacall-build.sh
+++ b/tools/metacall-build.sh
@@ -91,7 +91,7 @@ sub_build() {
 
 	# Memcheck
 	if [ $BUILD_MEMCHECK = 1 ]; then
-		ctest -j$(getconf _NPROCESSORS_ONLN) --timeout 5400 -T memcheck --output-on-failure
+		cmake --build . --target memcheck
 	fi
 
 	# Install


viferga:
Wow the logs are huge, I have found this for Ruby:
https://github.com/Shopify/ruby_memcheck/blob/main/suppressions/ruby.supp

viferga:
There's also a way to avoid false positives:

https://pythonextensionpatterns.readthedocs.io/en/latest/debugging/valgrind.html

I always thought to setup something like this at some point.

MrSpideyNihal:
> There's also a way to avoid false positives:
> 
> https://pythonextensionpatterns.readthedocs.io/en/latest/debugging/valgrind.html
> 
> I always thought to setup something like this at some point.



I generated it by reading the Valgrind output and manually identifying which errors came from third-party libraries (GTest, OpenSSL, libcurl, backward-cpp). 
I'll look at the Shopify ruby.supp and Python false positives approach you shared 
 looks like a better standard method. Should I update it using those?

viferga:
Create another PR with suppression files only, right now the error list is extreme.. XD

viferga:
I think we can close this PR once you push the suppression files.

MrSpideyNihal:
Okay

viferga:
Closing this in favor of other PRs.

PR #714, opened 2026-03-24
rs_loader: check for overflow on i64 to c_long cast

RaicyAugusto:
# Description

c_long is 32-bit on Windows. The previous cast from i64 to c_long was unchecked, causing silent overflow for values outside c_long range.
Added bounds check before casting and return Err on overflow.

## Type of change

# Checklist:


viferga:
it doesn't solve the issue in win

PR #722, opened 2026-03-25
fix(memcheck): add ruby suppression and remove duplicate wasm suppression

MrSpideyNihal:
## Changes

- Replaced duplicate `valgrind-wasm.supp` line with `valgrind-ruby.supp` in `cmake/CompileOptions.cmake`
- Added 3 missing suppressions to `valgrind-ruby.supp` from Shopify's ruby_memcheck reference file:
  - `strscan_do_scan` (regex leak in strscan.c)
  - `rb_callable_method_entry` (lazy callcache table creation)
  - `date__parse` (lazy Regexp initialization in date library)

## Context

`valgrind-ruby.supp` already existed in `source/tests/memcheck/` but was never referenced in `CompileOptions.cmake`. The duplicate `valgrind-wasm.supp` line on L88 was replaced with the ruby suppression.

Closes #704


viferga:
You pushed everything in the same old PR, don't do that, do this instead:

```
git checkout develop
git pull upstream develop 
git checkout -b fix/valgrind-supressions

// add the last changes you made

git push origin fix/valgrind-supressions

// create a pr from this branch into develop
```

PR #728, opened 2026-03-25
fix(rs_port): exit on missing library instead of silently warning

RaicyAugusto:
# Description

When `find_metacall_library()` failed, the `Err` branch only emitted a warning and still passed `cargo:rustc-link-lib=dylib=metacall`, letting the linker try an implicit fallback.

This could hide the real configuration issue because the build would continue even after explicit library detection had already failed.

Now it exits immediately with an error instead.

Before:
Screenshot from 2026-03-25 20-26-27

After:
image

## Type of change

### DIFF
--- a/source/ports/rs_port/sys/src/lib.rs
+++ b/source/ports/rs_port/sys/src/lib.rs
@@ -374,13 +374,8 @@ pub fn build() {
             }
             Err(e) => {
                 // Print the error
-                println!(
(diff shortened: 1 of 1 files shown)


viferga:
I have merged it but I am not sure if this is the best way, maybe we can return the result and exit(1) in rs_port instead of metacall-sys.

viferga:
Let's see what happens with the CI now, which CIs fail.

PR #731, opened 2026-03-26
Renamed pub fn build(source: Source) -> SourceImpl

### LINKED ISSUE #688: Enhancement: Renaming pub fn new(source: Source) -> SourceImpl
In the rs_loader, in the [compiler](https://github.com/metacall/core/tree/develop/source/loaders/rs_loader/rust/compiler) crates. In the `impl Source {...} ` there is a function new that returns a `SourceImpl`. In Rust, methods named `new` are semantically expected to return `Self` , so actually it's breaking rust convention and triggers `clippy::new_ret_no_self` lint.

I have temporally added  #[allow(clippy::new_ret_no_self)] in my PR #687, but I think that to have a more idiomatic code it should be renamed to build/create.

niko091:
# Description

Changed `pub fn build(source: Source) -> SourceImpl` to `fn build(source: Source) -> SourceImpl`

Removed #[allow(clippy::new_ret_no_self)] that now is useless

Changed the name every time the function is used

Updated the tests

Fixes #688 

## Type of change

# Checklist:

### DIFF
--- a/source/loaders/rs_loader/rust/compiler/src/file.rs
+++ b/source/loaders/rs_loader/rust/compiler/src/file.rs
@@ -12,7 +12,7 @@ pub struct FileRegistration {
 }
 impl FileRegistration {
     pub fn new(path_to_file: PathBuf) -> Result<FileRegistration, RegistrationError> {
-        let state = match compile(Source::new(Source::File {
+        let state = match compile(Source::build(Source::File {
             path: path_to_file.clone(),
         })) {
             Ok(state) => state,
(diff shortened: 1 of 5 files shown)


viferga:
why didn't you include this in the same PR of clippy refactor?

niko091:
Hi, because it was not a warning about code cleanliness, but a semantic  problem, so I created the Issue because I think that the #[allow(clippy::new_ret_no_self)] could be enough and because the Issue is still active I created this PR to maintain them separated. Let me know the preferred way to handle this.

niko091:
Closed in favor of #766

PR #622, opened 2026-02-04
Add Node.js 22+ support with Debian Forky

josead:
## Summary
- Change default base image from `debian:trixie-slim` to `debian:forky-slim` which provides Node.js 22+
- Add ICU 75.1 version mapping for Node.js 22+ in CMake configuration (for building Node.js from source)
- Update documentation to reflect supported Node.js versions (>= 18.x.x)
- Add `debian:forky-slim` to CI test matrix for both regular and sanitizer tests
- Remove discontinued Alpine Linux support from comments

## Test plan
- [ ] Run `./docker-compose.sh build` with `METACALL_BASE_IMAGE=debian:forky-slim`
- [ ] Run `./docker-compose.sh test` to verify all tests pass
- [ ] Verify Node.js version is 22+ in the container
- [ ] Check that LLVM/C loader works (Vicente mentioned potential issues)

## Files changed
- `.env` - Default base image updated
- `cmake/FindNodeJS.cmake` - ICU version for Node.js 22+
- `docs/README.md` - Updated version documentation
- `.github/workflows/linux-test.yml` - CI matrix updated

## Notes
- macOS already supports Node.js 22 via Homebrew and pre-built libnode binaries
- Windows support remains at Node.js 14.18.2 (out of scope for this PR)
- Alpine Linux support is discontinued

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #732, opened 2026-03-26
fix(environment): Add libnode-dev for Ubuntu/Debian to fix Node loader build

0xShubhamSolanki:
## Problem

On Ubuntu 24.04, the build fails because `libnode-dev` is not installed by the environment script. This causes CMake to fall back to building Node from source, which fails due to Python 3.12 incompatibility (distutils removal).

Without the development headers, `FindNodeJS.cmake` cannot detect the system Node, triggering a broken source build fallback.

## Solution

Add `libnode-dev` to the package list in `sub_nodejs()` for Debian/Ubuntu. This ensures Node development headers are available, allowing MetaCall to properly detect and link against the system Node.

## Manual Testing Performed

- Fresh Ubuntu 24.04 environment
- Ran `./tools/metacall-environment.sh base python nodejs rpc`
- Build completed successfully with all loaders
- Verified CLI with Python and Node loaders working correctly
- Verified with `dpkg -l | grep libnode-dev`

## Context

Discussed in Discord #general channel with community members facing the same issue on Ubuntu 24.

---

## Type of change

- [ ] Documentation update

## Checklist:


viferga:
Provide a test for reproducing the issue before sending the patch.

viferga:
Also, is it related to this?
https://github.com/metacall/core/pull/622

0xShubhamSolanki:
Hey, thanks for pointing me to #622 , i went through it. From what i have understood, it is mainly about adding Node.js 22+ support with newer Debian images (forky/trixie), along with some Cmake and CI updates.

My PR (#732) is a bit different .it is mainly focused on fixing the current build issue on ubuntu 24.04. Right now, the environment script doesn’t install libnode-dev, so CMake can’t detect the system Node.js and falls back to building from source, which then fails due to python 3.12 (distutils removal).

From what I can tell, #622 focuses on Docker images and newer Node, but doesn't add the `libnode-dev` package to the environment script. So #732 would still help Ubuntu 24 users running outside Docker. So while both touch Node.js setup, this one is more of a small fix for the current Ubuntu setup rather than adding support for newer Node versions.  

For the test part i have added a verification check inside the environment script itself,the test run right after the package installation step, so it catches the issue early. Does this approach work? Or would you prefer a separate reproduction script?

PR #735, opened 2026-03-27
Fix macOS library discovery in "next" metacall-sys 0.1.5

k5602:
## Summary
This PR fixes two bugs in `metacall-sys` that affect Cargo-based MetaCall consumers on macOS and in CI environments.

### Root Cause

The investigation started from repeated macOS CI failures in metassr and traced the problem back to `metacall-sys` behavior.

Two issues were identified:

- `METACALL_INSTALL_PATH=""` was treated as a real custom path instead of being ignored
- macOS library discovery needed regression coverage to guarantee both release and debug dylib names remain searchable

In affected consumers, this could cause library discovery to fail and eventually fall back to linking `-lmetacall` even when only the debug dylib was installed.

### Changes

- ignore empty `METACALL_INSTALL_PATH` values and fall back to the normal platform-specific search paths
- add regression tests covering:
  - empty custom install path fallback
  - non-empty custom install path override
  - macOS release/debug dylib name matching
- bump `metacall-sys` version from `0.1.4` to `0.1.5`
- update the local `metacall` Rust port to depend on `metacall-sys 0.1.5`

### Validation

- `cargo test -p metacall-sys`
- `cargo check -p metacall-sys`

### Impact

This release unblocks downstream Cargo consumers that rely on explicit library path injection in CI, especially on macOS where only the debug MetaCall dylib may be installed so after releasing and re-evaluating we can clean the CI.

### DIFF
--- a/source/ports/rs_port/Cargo.toml
+++ b/source/ports/rs_port/Cargo.toml
@@ -19,7 +19,7 @@ path = "src/lib.rs"
 metacall-inline = { version = "0.2.0", path = "./inline" }
 
 [build-dependencies]
-metacall-sys = { version = "0.1.4", path = "./sys" }
+metacall-sys = { version = "0.1.5", path = "./sys" }
 bindgen = "0.72.1"
 
 [workspace]
--- a/source/ports/rs_port/sys/Cargo.toml
+++ b/source/ports/rs_port/sys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "metacall-sys"
-version = "0.1.4"
+version = "0.1.5"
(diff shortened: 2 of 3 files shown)


viferga:
I don't think this is the issue, we have updated metacall-sys recently, I think changes were just not propagated: https://github.com/metacall/core/actions/runs/24051937318

viferga:
https://github.com/metacall/metassr/pull/137/changes#diff-107e910e9f2ebfb9a741fa10b2aa7100cc1fc4f5f3aca2dfe78b905cbd73c0d2

PR #736, opened 2026-03-27
feat(msan): add clang install and memory sanitizer configuration

MrSpideyNihal:
### tools/metacall-environment.sh
- Added `INSTALL_CLANG=0` flag
- Added `sub_clang()` function that installs clang via apt on Linux
- Added clang option to `sub_options()` and `sub_install()`
- Added clang to help text

### tools/metacall-configure.sh
- Added `BUILD_MEMORY_SANITIZER=0` flag
- Added memory-sanitizer option to `sub_options()`
- Sets `CC=clang` and `CXX=clang++` as env vars before cmake (not as -D flags)
- Enables `-DOPTION_BUILD_MEMORY_SANITIZER=On -DOPTION_BUILD_ADDRESS_SANITIZER=Off`
- Added memory-sanitizer to help text

## Notes

Compiler is set via `CC`/`CXX` env vars instead of `-DCMAKE_C_COMPILER` flags to follow the same pattern as existing sanitizers and avoid breaking the BUILD_STRING line-by-line parsing.

CI job will be added in a follow-up as discussed with mentor.


tools/metacall-configure.sh
```diff
@@ -546,7 +551,7 @@ sub_configure() {
 
 	# Memory Sanitizer
 	if [ $BUILD_MEMORY_SANITIZER = 1 ]; then
-		BUILD_STRING="$BUILD_STRING -DOPTION_BUILD_MEMORY_SANITIZER=On"
+		BUILD_STRING="$BUILD_STRING -DOPTION_BUILD_MEMORY_SANITIZER=On -DOPTION_BUILD_ADDRESS_SANITIZER=Off"
```
viferga: undo this change

tools/metacall-configure.sh
```diff
@@ -564,7 +569,11 @@ sub_configure() {
 	BUILD_STRING="$BUILD_STRING -DCMAKE_BUILD_TYPE=$BUILD_TYPE"
 	
 	# Execute CMake
-	cmake -Wno-dev -DOPTION_GIT_HOOKS=Off $BUILD_STRING ..
+	if [ $BUILD_CLANG = 1 ]; then
+		CC=clang CXX=clang++ cmake -Wno-dev -DOPTION_GIT_HOOKS=Off $BUILD_STRING ..
```
viferga: here just export the env vars, don't duplicate the cmake command

if ..
export ...
endif 

cmake ...

docker-compose.sh
```diff
@@ -197,6 +197,30 @@ sub_test_memcheck() {
 	$DOCKER_COMPOSE -f docker-compose.yml -f docker-compose.test.yml build --force-rm dev
 }
 
+# Build MetaCall Docker Compose with Clang for testing (link manually dockerignore files)
+sub_test_clang() {
+	# Disable BuildKit as workaround due to log limits (TODO: https://github.com/docker/buildx/issues/484)
+	export DOCKER_BUILDKIT=0
+
+	# Enable build with clang
+	export METACALL_BUILD_SANITIZER=clang
```
viferga: this is wrong, maybe we should add a new enviroment variable for clang, but the tests will work using this, I'm gonna to run the ci

MrSpideyNihal: yes , will add METACALL_BUILD_CLANG as a separate env var once CI runs

PR #737, opened 2026-03-27
test: add ConcurrentSyncInvoke test to reproduce invoke_curl data rac…

### LINKED ISSUE #693: function_rpc_interface_invoke  is not thread-safe  (shared invoke_curl handle causes data race)
### Expected Behavior

Concurrent RPC function invocations from multiple threads should execute independently without interfering with each other's request parameters.

### Current Behavior

All synchronous invocations share a single CURL *invoke_curl handle ([rpc_loader_impl.cpp:68](https://github.com/metacall/core/blob/develop/source/loaders/rpc_loader/source/rpc_loader_impl.cpp#L68)). The setopt + perform sequence in [function_rpc_interface_invoke (l200-205) ](https://github.com/metacall/core/blob/develop/source/loaders/rpc_loader/source/rpc_loader_impl.cpp#L200-L205 ) is not atomic.

If two threads interleave there, then one thread's URL can end up paired with the other thread's post body which might send the wrong payload to the wrong endpoint.

SOmething I found interesting is that ([function_rpc_interface_await L386](https://github.com/metacall/core/blob/develop/source/loaders/rpc_loader/source/rpc_loader_impl.cpp#L386)) already creates a per-call CURL * easy handle, so the async path does not have this problem.

Also, while looking at this function, I noticed the error log on [L212](https://github.com/metacall/core/blob/develop/source/loaders/rpc_loader/source/rpc_loader_impl.cpp#L212) uses [%] as a format specifier. Shouldn't that be [%s]? The async path on [L277](https://github.com/metacall/core/blob/develop/source/loaders/rpc_loader/source/rpc_loader_impl.cpp#L277) uses %s.

### Possible Solution

Mirror what  `function_rpc_interface_await` already does i.e. create a fresh CURL * easy handle per invocation instead of reusing the shared invoke_curl one. This eliminates the race entirely without needing a mutex.

### Steps to Reproduce

1. Initialize the RPC loader and discover functions from a remote endpoint.
2. From two or more threads, concurrently invoke different RPC functions
3. Observe that requests may be sent to the wrong URL, with the wrong body or crash due to concurrent mutation of the shared CURL * handle's internal state.

### Context (Environment)

This came up while I was reviewing the RPC loader code in the for the Function Mesh GSoC project. In a mesh topology, multiple languages and threads will be invoking different RPC endpoints concurrently,  this race condition is a blocker for that use case. The async path (await) is already safe since it allocates per-call handles, so this inconsistency between `invoke`  and  `await `doesn't look like it was ntentional.

### Detailed Description

The ([loader_impl_rpc_type struct](https://github.com/metacall/core/blob/develop/source/loaders/rpc_loader/source/rpc_loader_impl.cpp#L536)) holds a single `CURL *invoke_curl` that is initialized once during rpc_loader_impl_initialize ([L536](https://github.com/metacall/core/blob/develop/source/loaders/rpc_loader/source/rpc_loader_impl.cpp#L536)) and then reused by every synchronous call. The `CURLOPT_URL`, `CURLOPT_POSTFIELDS`, `CURLOPT_POSTFIELDSIZE`, and `CURLOPT_WRITEDATA` options are all set on this shared handle immediately before `curl_easy_perform`. 

The `discover_curl` handle has the same pattern but is less likely to be called concurrently since discovery typically happens once at load time.

### Possible Implementation

Replace the shared `nvoke_curl` approach with per-call handles.

Codingisinmyblud:
# Description

I added a test that spawns 4 threads making concurrent synchronous metacallht_s('divide') calls, which would then trigger the data race on the shared invoke_curl handle in function_rpc_interface_invoke.

Fixes #693 

## Type of change

- [] Bug fix (non-breaking change which fixes an issue)

# Checklist:

- [x] If my change is significant or breaking, I have passed all tests with `./docker-compose.sh test &> output` and attached the output.
- [x] I have tested my code with `OPTION_BUILD_ADDRESS_SANITIZER` or `./docker-compose.sh test-address-sanitizer &> output` and `OPTION_TEST_MEMORYCHECK`.
- [x] I have tested my code with `OPTION_BUILD_THREAD_SANITIZER` or `./docker-compose.sh test-thread-sanitizer &> output`.
- [x] I have tested with `Helgrind` in case my code works with threading.

### DIFF
--- a/source/loaders/rpc_loader/source/rpc_loader_impl.cpp
+++ b/source/loaders/rpc_loader/source/rpc_loader_impl.cpp
@@ -65,7 +65,6 @@ struct rpc_async_context;
 typedef struct loader_impl_rpc_type
 {
 	CURL *discover_curl;
-	CURL *invoke_curl;
 	CURLM *async_multi;
 	std::thread poll_thread;
 	std::atomic<bool> exit_flag;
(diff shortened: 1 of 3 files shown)


Codingisinmyblud:
With the new tests added, I built with ThreadSanitizer and ran the new ConcurrentSyncInvoke test and outputted the results in a txt file to clearly see the data race.
Here's a screenshot to see the data race:

image

viferga:
this is the test only without the fix, right?

viferga:
I'm running the CI to trigger the issue.

viferga:
It worked, tests are failing with data races and heap-use-after-free, if you PR the fix I will merge it.

Codingisinmyblud:
Thanks, I am working on the fix now, I'll commit the changes soon hopefully.

viferga:
@anasbarg finally updated it: https://github.com/metacall/core/pull/765

I am closing this one in favor of the other, but thank you so much for pointing out the issue.

PR #738, opened 2026-03-27
Automated Ruby Gem release pipeline and Cross Platform CI fixes 

avik22835:
Hi Vicente,

The goal of this PR is to enable a fully automated release pipeline for the Metacall Ruby port.

I've implemented a hybrid workflow (release-ruby.yml) and a helper script (upload.sh) that handles everything from environment detection to the final push to RubyGems.org.

The Release pipeline logic

1. upload.sh 
  * Version Parsing: It uses a Ruby one-liner (ruby -e) to load the .gemspec and extract the local version.
  * Remote Check: Before building, it uses curl to check the RubyGems API. If the current version already exists on the registry, the script exits silently.
  * Secure Auth: It uses the RUBYGEMS_API_KEY to temporarily create a credentials file for the push.

2. release-ruby.yml
  * Job 1: Test: It runs the integration suite across ubuntu-latest, macos-latest, and windows-latest. I've solved the fragility issues within them.
  * Job 2: Release: This job is locked until its needed. It only fires when a version tag (like v0.1.0) is pushed or manually triggered via workflow dispatch. It requires the test job to pass first.
  
Solving the Windows "Error 126". I've fixed this by:
  * Standardizing on the bundled Metacall Ruby 3.5 runtime to ensure binary compatibility.
  * Fixing a UTF-8/BOM encoding issue where Powershell was corrupting the environment file, making it unreadable to the Github Runner.
  
Current Status and Evidence
I have verified this entire flow on my fork. As you can in the attached screenshot, the integration tests are now 3/3 green across the entire matrix, and the release job currently says skipped cause no version tags have been pushed.
WhatsApp Image 2026-03-27 at 13 04 46

To enable the automated gem push, please add the following Secret to the metacall/core repository:
  * Name: RUBYGEMS_API_KEY
  * Value: Your API key from RubyGems.org with "Push Gem" permissions.

### DIFF
--- a/source/ports/rb_port/package/lib/metacall.rb
+++ b/source/ports/rb_port/package/lib/metacall.rb
@@ -25,14 +25,12 @@ module MetaCall
 
 	def platform_install_paths
 		host_os = RbConfig::CONFIG['host_os']
-		home_dir = Dir.home
 
 		case host_os
 		when /mswin|mingw|cygwin/
 			{
 				paths: [ 
 					File.join(ENV['LOCALAPPDATA'].to_s, 'MetaCall', 'metacall'),
-					File.join(home_dir, 'AppData', 'Local', 'MetaCall', 'metacall')
 				],
 				name: '^metacall(d)?\.dll$'
 			}
@@ -41,8 +39,6 @@ module MetaCall
 				paths: [ 
 					'/opt/homebrew/lib/', 
 					'/usr/local/lib/', 
-					File.join(home_dir, '.metacall', 'lib'),
-					'/opt/metacall/lib'
 				],
 				name: '^libmetacall(d)?\.dylib$'
 			}
@@ -51,8 +47,6 @@ module MetaCall
 				paths: [ 
 					'/usr/local/lib/', 
 					'/gnu/lib/', 
-					File.join(home_dir, '.metacall', 'lib'),
-					'/opt/metacall/lib'
 				],
 				name: '^libmetacall(d)?\.so(\.\d+)*$'
 			}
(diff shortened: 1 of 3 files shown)


viferga:
It looks pretty good but the algorithm for searching the library must be in ruby itself, not in bash or ps1:

Nodejs:
https://github.com/metacall/core/blob/70362ecf24ab4f377761e75da3febc1bac02feb0/source/ports/node_port/index.js#L28-L159

Python:

https://github.com/metacall/core/blob/70362ecf24ab4f377761e75da3febc1bac02feb0/source/ports/py_port/metacall/api.py#L26-L112

source/ports/rb_port/package/lib/metacall.rb
```diff
@@ -24,22 +25,39 @@ def find_files_recursively(root_dir, pattern)
...
+				name: '^metacall(d)?\.dll$'
 			}
 		when /darwin/
 			{
-				paths: [ '/opt/homebrew/lib/', '/usr/local/lib/' ],
-				name: 'libmetacall\.dylib'
+				paths: [ 
+					'/opt/homebrew/lib/', 
+					'/usr/local/lib/', 
+					File.join(home_dir, '.metacall', 'lib'),
```
viferga: why those paths?

source/ports/rb_port/package/lib/metacall.rb
```diff
@@ -84,11 +102,39 @@ def metacall_module_load
...
-
-		# Find and load the MetaCall shared library
+		# Find the MetaCall shared library
 		library_path = find_library
+		install_dir = File.dirname(library_path)
+		root_dir = File.dirname(install_dir)
+
+		# Set environment variable for the host
+		ENV['METACALL_HOST'] ||= 'rb'
+
```
viferga: all those env vars in theory are setup automatically, we should check if it works after deleting them

source/ports/rb_port/package/lib/metacall.rb
```diff
@@ -84,11 +102,39 @@ def metacall_module_load
...
-		ENV['METACALL_HOST'] = 'rb'
-
-		# Find and load the MetaCall shared library
+		# Find the MetaCall shared library
 		library_path = find_library
+		install_dir = File.dirname(library_path)
+		root_dir = File.dirname(install_dir)
+
+		# Set environment variable for the host
+		ENV['METACALL_HOST'] ||= 'rb'
```
viferga: this one is correct

source/ports/rb_port/package/lib/metacall.rb
```diff
@@ -84,11 +102,39 @@ def metacall_module_load
...
+		# This eliminates the need for manual "Smart CI" scripts.
+		ENV['LOADER_LIBRARY_PATH'] ||= install_dir
+		ENV['SERIAL_LIBRARY_PATH'] ||= install_dir
+		ENV['DETECTOR_LIBRARY_PATH'] ||= install_dir
+		
+		# Look for configurations folder (usually adjacent to lib/bin in self-contained)
+		config_path = File.join(root_dir, 'configurations')
+		ENV['CONFIGURATION_PATH'] ||= config_path if Dir.exist?(config_path)
+
+		# Platform-specific environment fixes
```
viferga, line 102: respect to this, I've seen the issue twice...

maybe we should solve it externally with something like what linux distributable does, it provides a file for defining all env vars before hand that you can use to load them...

although I'm not sure what is the best method, if we do it here we will need to replicate it everywhere 


also I think this issue also comes because we are relocating the runtimes in windows, maybe this is the main issue

avik22835:
So my PR is now clean. All env related parts are removed while keeping the matching logic intact. 
Is it ready for merge?

source/ports/rb_port/package/lib/metacall.rb
```diff
@@ -84,11 +99,51 @@ def metacall_module_load
...
+				
+				# Add the Ruby runtime bin folder if it exists
+				rb_runtime_bin = File.join(root_dir, 'runtimes', 'ruby', 'bin')
+				set_dll_dir.call(rb_runtime_bin.encode('UTF-16LE')) if Dir.exist?(rb_runtime_bin)
+			rescue => e
+				# Fallback to PATH for older Ruby versions
+				ENV['PATH'] = "#{install_dir};#{ENV['PATH']}"
+			end
+			
+			# Detect Python runtime bundled with MetaCall
```
viferga: this shouldn't go here, we were talking with @k5602 to implement it into the loader directly or in some common site

viferga:
The last run seems to still have failed, is this correct?
https://github.com/metacall/core/actions/runs/23642751953/job/68870397210?pr=738

PR #740, opened 2026-03-27
Migrate rs_loader to nightly-2024-11-01

Harsh-dev023:
## What this does

Migrates the `rs_loader` from `nightly-2021-12-04` to `nightly-2024-11-01` — a ~3 year jump in `rustc` internals.

The old pinned nightly was blocking us from picking up compiler API improvements and was increasingly out of step with the ecosystem. This updates all call sites that depend on `rustc` internals.

## Changes (one file per commit for reviewability)

1. **`rust-toolchain`** — bump to `nightly-2024-11-01`
2. **`lib.rs`** — restructure compiler callbacks (`after_crate_root_parsing` / `after_expansion`), update `Config` struct, replace deprecated std types (`SyncLazy` → `LazyLock`, `PanicInfo` → `PanicHookInfo`), drop stabilized feature gates, simplify ICE reporting
3. **`middle.rs`** — `TyS` → `Ty<'tcx>`, unwrap `EarlyBinder` from `fn_sig()`/`type_of()`, `Export` → `ModChild`, add `f16`/`f128` fallback
4. **`ast.rs`** — `TyKind::Rptr` → `Ref`, `VariantData::Struct` pattern update, wildcard arms for new variants
5. **`wrapper/mod.rs`** — fix `FileName` import path
6. **`wrapper/class.rs`** — convert `i64` → `c_long` before FFI call (Windows `c_long` = `i32`)

## Testing

The public API of `rs_loader` is unchanged — only internal `rustc` call sites were updated. Should be verified against the existing test suite on the CI nightly toolchain. Compiles cleanly against `nightly-2024-11-01`.


viferga:
Is there an specific reason you used 2024-11-01? We are on 2026.

Harsh-dev023:
Yeah fair point! The original was on `nightly-2021-12-04` so I picked `2024-11-01` as a safe middle ground — rs_loader uses rustc's private internals which break a lot between nightlies. Didn't want to bite off too much at once. Happy to bump it further if you'd like!

PR #743, opened 2026-03-28
ci: add linux memory sanitizer test job

MrSpideyNihal:
### docker-compose.sh
- Added `sub_test_memory_sanitizer()` — sets METACALL_BUILD_CLANG=clang and METACALL_BUILD_SANITIZER=memory-sanitizer
- Updated `test-memory-sanitizer` case to call the new dedicated function instead of sub_test_sanitizer

### .github/workflows/linux-test.yml
- Added `linux-memory-sanitizer-test` job following the same pattern as `linux-clang-test`

## Notes
Clang is installed via METACALL_BUILD_CLANG which is already wired in docker-compose.test.yml from the previous PR.

### DIFF
--- a/cmake/CompileOptions.cmake
+++ b/cmake/CompileOptions.cmake
@@ -119,9 +119,10 @@ if(OPTION_BUILD_THREAD_SANITIZER AND (CMAKE_BUILD_TYPE STREQUAL "Debug" OR CMAKE
 		"__THREAD_SANITIZER__=1"
 	)
 elseif(OPTION_BUILD_MEMORY_SANITIZER AND "${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang" AND (CMAKE_BUILD_TYPE STREQUAL "Debug" OR CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo"))
-	# TODO: This requires much more effort than expected: https://github.com/google/sanitizers/wiki/MemorySanitizerLibcxxHowTo
-	set(SANITIZER_LIBRARIES)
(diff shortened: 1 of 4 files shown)


viferga:
It looks very well, but it's not ready. We don't have symbols and they are required for debugging.

Try to solve this issue:
```
==10203==WARNING: MemorySanitizer: use-of-uninitialized-value
==10203==WARNING: invalid path to external symbolizer!
```

@MrSpideyNihal

cmake/CompileOptions.cmake
```diff
@@ -407,7 +408,7 @@ if (PROJECT_OS_FAMILY MATCHES "unix" OR PROJECT_OS_FAMILY MATCHES "macos")
 		add_compile_options(-fsanitize=memory)
 		add_compile_options(-fsanitize-memory-track-origins)
 		add_compile_options(-fsanitize-memory-use-after-dtor)
-		if(PROJECT_OS_FAMILY MATCHES "macos")
+		if(PROJECT_OS_FAMILY MATCHES "macos" OR PROJECT_OS_FAMILY MATCHES "unix")
```
viferga: should we delete the condition entirely?

viferga:
The code goods good in general.

cmake/CompileOptions.cmake
MrSpideyNihal: yes, will remove the OS condition entirely and fix the symbolizer issue by installing llvm-symbolizer and adding the path to MSAN_OPTIONS

viferga:
I have merged it but there's still a task to do.

PR #747, opened 2026-03-30
ci: add cargo fmt --check to Rust CI pipeline

Harsh-dev023:
## What
Added a `fmt` job to the Rust CI pipeline that runs `cargo fmt --check` before the build and test steps.

## Why
Formatting was not being enforced in the Rust CI. `cargo clippy` was already present for linting but there was no check for code formatting.

## Changes
- Added a new `fmt` job in `.github/workflows/release-rust.yml`
- Runs `cargo fmt --check` on `source/ports/rs_port`
- Added `needs: fmt` to the `test` job so the pipeline order is: `fmt → test → release`

## Notes
- `--check` flag is used so CI fails if code is not formatted, without modifying files
- `fmt` job only runs on `ubuntu-latest` since formatting is OS-independent


viferga:
Do it in the same job

Harsh-dev023:
> Do it in the same job

Got it sir, will move the cargo fmt --check step into the existing test job instead of a separate job. Pushing the fix now!

viferga:
still failing, ignore bindings.rs

fahdfady (commented):
I think that requires that we have a `rustfmt.toml` first. because formatting might be different from one contributor to another depending on toolchain and version.
e.g. I use nightly and another one uses stable. that might produce conflict while we both have done `cargo fmt` before we push our code.

PR #751, opened 2026-03-30
fix(ext_loader): add default constructors to prevent uninitialized memory

MrSpideyNihal:
## Summary

Fix uninitialized memory detected by MemorySanitizer in ext_loader_impl_execution_path.

## Changes

Added default constructors to `loader_impl_ext_type` and `loader_impl_ext_handle_type` structs to explicitly initialize their members (`paths`, `destroy_list`, `extensions`).

## Context

This bug was found while implementing the Memory Sanitizer CI job. MSan reported use-of-uninitialized-value repeatedly at ext_loader_impl.cpp:136.


viferga:
Apparently this doesn't solve the issue. You should test locally before PRing.

PR #762, opened 2026-04-08
Add freebsd ci

yug105:
Starting FreeBSD CI support for metacall/core.                                                                                                     
  - Added FreeBSD OS detection to environment and configure scripts
  - Added base package installation via pkg
  - Added freebsd-test.yml using cross-platform-actions for FreeBSD 14.2

### DIFF
--- a/cmake/FindNodeJS.cmake
+++ b/cmake/FindNodeJS.cmake
@@ -39,12 +39,6 @@ if(NodeJS_EXECUTABLE)
 	set(NodeJS_FIND_QUIETLY TRUE)
 endif()
 
-if(PROJECT_OS_BSD)
-    set(MAKE_COMMAND gmake)
-else()
-    set(MAKE_COMMAND make)
-endif()
-
 option(NodeJS_CMAKE_DEBUG "Print paths for debugging NodeJS dependencies." OFF)
 option(NodeJS_SHARED_UV "If it is enabled, libuv won't be required by this script." OFF)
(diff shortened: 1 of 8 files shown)


cmake/FindNodeJS.cmake
```diff
@@ -425,7 +425,13 @@ if(NOT NodeJS_LIBRARY)
 		message(STATUS "Extract NodeJS distribution")
 		execute_process(COMMAND ${CMAKE_COMMAND} -E tar "xvf" "${NodeJS_DOWNLOAD_FILE}" WORKING_DIRECTORY "${NodeJS_BASE_PATH}" OUTPUT_QUIET)
 	endif()
-
+	if(PROJECT_OS_BSD)
+		execute_process(
+			COMMAND sed -i '' "/ASSERT_TRIVIALLY_COPYABLE(T);/d" "${NodeJS_OUTPUT_PATH}/deps/v8/src/base/small-vector.h"
+			WORKING_DIRECTORY "${NodeJS_OUTPUT_PATH}"
```
viferga: What's this, aren't you building it now from metacall-environment.sh instead no?

yug105: Yes you're right, since we build Node in environment.sh now, FindNodeJS.cmake finds libnode.so already installed and skips the source build. The patch there was leftover from before I moved the build to environment.sh. Sorry for not cleaning it up sir.

PR #763, opened 2026-04-10
chore: add tools/instrumentation folder with MSan and python-valgrind scripts

MrSpideyNihal:
## Changes

Creates `tools/instrumentation/` as a dedicated home for opt-in instrumentation helpers, as discussed with mentor.

### Structure
- `python-valgrind/` — build script for Python with Valgrind support (`--with-pydebug --without-pymalloc --with-valgrind`)
- `msan/` — build script for LLVM/libc++/gtest with MSan instrumentation
- `suppressions/` — aggressive suppression files used only for local deep triage (not wired into default CI)

### Notes
- Aggressive `obj:*libruby*` suppressions moved here from main suppression files
- Scripts not consumed by default CI unless explicitly wired
- Future: cmake/CompileOptions.cmake will be updated to use these flows


viferga:
Do you have the dockerfile for the msan?

MrSpideyNihal:
> Do you have the dockerfile for the msan?

haven't run the MSan Dockerfile end-to-end yet 
 building LLVM can take a wile.
 it's a reference implementation based on the wiki the python-valgrind one is the tested one.

Harsh-dev023:
> > Do you have the dockerfile for the msan?
> 
> haven't run the MSan Dockerfile end-to-end yet building LLVM can take a wile. it's a reference implementation based on the wiki the python-valgrind one is the tested one.

are we adding valgrid in python itself??
sound's cool

viferga:
> > > Do you have the dockerfile for the msan?
> > 
> > 
> > haven't run the MSan Dockerfile end-to-end yet building LLVM can take a wile. it's a reference implementation based on the wiki the python-valgrind one is the tested one.
> 
> are we adding valgrid in python itself?? sound's cool

yep that's exactly what we are doing 🗿

PR #772, opened 2026-04-16
BSD support without nodejs 

yug105:
- Disables NodeJS temporarily for faster CI iteration
- And adding more languages to loader
- Adding more architectures
- fixing the release for dylink and detour test


tools/metacall-environment.sh
```diff
@@ -114,7 +114,7 @@ sub_base(){
 	elif [ "${OPERATIVE_SYSTEM}" = "Darwin" ]; then
 		brew install llvm cmake git wget gnupg ca-certificates
 	elif [ "${OPERATIVE_SYSTEM}" = "FreeBSD" ]; then
-		$SUDO_CMD pkg install -y cmake git gmake wget gnupg ca_root_nss
+		$SUDO_CMD pkg install -y cmake git gmake wget gnupg ca_root_nss libunwind libdwarf libelf
```
viferga: this goes into sub_backtrace

source/tests/dynlink_test/CMakeLists.txt
```diff
@@ -133,6 +133,7 @@ target_link_options(${target}
 	# Required for linking symbols from the executable
 	$<$<AND:$<PLATFORM_ID:Darwin>,$<CXX_COMPILER_ID:AppleClang,Clang>>:-Wl,-export_dynamic>
 	$<$<CXX_COMPILER_ID:GNU>:-rdynamic>
+	$<$<AND:$<PLATFORM_ID:FreeBSD>,$<CXX_COMPILER_ID:Clang>>:-Wl,--export-dynamic>
```
viferga, line 136: what compiler are you using in freebsd?

viferga: @yug105 you added here clang but are we using clang? is this flag correct?

yug105: yes it is correct

yug105: we are using clang

viferga: I think gcc also works so we can remove the requirement of being clang, because both compilers support this flag.

viferga: Oh sorry it's correct, we are covering it with -rdynamic before.

Issue #746, opened 2026-03-30
macOS: dynlink unix path resolves .so, but c_loader package load needs .dylib (libgit2)


ROKUMATE:
On macOS, C package loading fails because dynlink follows the unix implementation path that resolves shared libraries as `.so`, while Homebrew/system `libgit2` is provided as `.dylib`.

This breaks C loader package resolution and causes macOS CI failures in the C loader path.

Basically what happens is that dynlink is being used by c_loader to load libraries from package, but dynlink in macos only supports .so under posix implementation, it supports dynlink but with an old implementation, the system is not designed to support multiple extensions by one implementation then when it tried to load a .dylib,  it fails.

### Problem Identified 
- dynlink unix flow resolves `.so`
- macOS libgit2 artifact is `.dylib`
- c_loader uses dynlink for package loading
- result: library is not loaded, downstream symbol discovery/load fails

### Failing test reference
- `metacall-node-port-c-lib-test`

### Failing CI logs
- Workflow run: https://github.com/metacall/core/actions/runs/22696380796
- Example failing job: https://github.com/metacall/core/actions/runs/23293047971/job/67737711135

PR #773, opened 2026-04-16
fixed the mac-ci failing for metacall_node_port_c_lib_test

ROKUMATE:
## Summary

Export `metacall_load_from_package_ex` to the node_port layer so C libraries can be loaded with all options (include paths, headers, libraries) in a single call. Fixes macOS CI failure where C loader couldn't be properly initialized through node_port.

## Testing
- tested on mac - 26.3.1 locally
- ✅ 74/74 comprehensive tests passing
- ✅ `metacall-node-port-c-lib-test` passing (critical test)

## Details

The C loader requires `headers` option to discover and export symbols via libclang. Previous approach using `metacall_execution_path` was fragile on macOS. New approach passes all options at load time, ensuring symbols are properly discovered.

## Closes
- #746

### DIFF
--- a/source/loaders/node_loader/source/node_loader_port.cpp
+++ b/source/loaders/node_loader/source/node_loader_port.cpp
@@ -886,6 +886,95 @@ napi_value node_loader_port_metacall_load_from_package_export(napi_env env, napi
 	return v_exports;
 }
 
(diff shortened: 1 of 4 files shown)


viferga:
You made it, congratulations. It passes the macos tests.

PR #778, opened 2026-04-20
Bsd implementation

yug105:


viferga:
I think we should just leave arm64 in freebsd as TODO, then test with everything to see if it works and solve the release bugs. That should be enough.

The detours bug in release may be due to having the symbols available due to the optimizer removing them out, it's a common thing but we must check in detail.

tools/metacall-environment.sh
```diff
@@ -616,6 +616,8 @@ sub_rpc(){
 		fi
 	elif [ "${OPERATIVE_SYSTEM}" = "Darwin" ]; then
 		brew install curl
+	elif [ "${OPERATIVE_SYSTEM}" = "FreeBSD" ]; then
```
viferga, line 619: This is the only thing I would maintain from this PR.

viferga:
Didn't you pull from upstream?

yug105:
> Didn't you pull from upstream?

Yes, I've rebased on latest develop now. The branch was out of sync before.

PR #800, opened 2026-05-24
Add metacall_loader_types API 

ROKUMATE:
- Exposes the per-loader (name - type_id) mapping through the public C API
so meta-ast / stub gen / function mesh share a single source of truth
instead of duplicating normalization.

- API:
```c
  struct metacall_loader_type { const char *name; enum metacall_value_id id; };
  int metacall_loader_types(const char *tag,
                            struct metacall_loader_type **out_array,
                            size_t *size);
```

Caller frees the array with free; names are loader-owned. Loader must
already be initialized.

- Implementation: new `loader_impl_type_count` / `loader_impl_type_iterate`
walk the existing `type_info_map` populated at loader init — no new
registration path.

- Test: `source/tests/metacall_loader_types_test/`
- Local CI: https://github.com/ROKUMATE/MetaCall-core/actions/runs/26357987907/job/77588143042#step:3:17433


viferga:
It looks good but avoid callbacks if you can, we can use iterators instead.

ROKUMATE:
is this the iterator shape you wanted, mirroring set_iterator_type?
```c
struct loader_impl_type_iterator_type
{
    struct set_iterator_type it;
};
typedef struct loader_impl_type_iterator_type *loader_impl_type_iterator;

LOADER_API void loader_impl_type_iterator_begin(loader_impl_type_iterator it, loader_impl impl);
LOADER_API int  loader_impl_type_iterator_end(loader_impl_type_iterator it);
LOADER_API void loader_impl_type_iterator_next(loader_impl_type_iterator it);
LOADER_API const char *loader_impl_type_iterator_name(loader_impl_type_iterator it);
LOADER_API type_id loader_impl_type_iterator_id(loader_impl_type_iterator it);
```
and which header ... `loader_impl.h` (alongside the functions, like `adt_set.h` does) or `loader_impl_interface.h` (next to the `loader_impl typedef`) ??

viferga:
Here's the review: https://github.com/metacall/core/pull/801

PR #810, opened 2026-06-08
feat: add Python type conversion test

Harsh-dev023:
## What
Added a type conversion test for the Python loader, similar to the existing Node.js conversion test.

## Why
To check what type conversions happen when calling an identity function through the Python loader.

## Changes
- Added source/tests/metacall_py_conversion_test/ with test source and CMakeLists.txt
- Added the new test to source/tests/CMakeLists.txt

## Notes
- Python int maps to long (METACALL_LONG) unlike Node.js which maps to double
- Added TODO comment to loop through all types once viferga pushes his full changes


ROKUMATE:
hey harsh, been working on the same issue 
i have explored that the CLASS, OBJECT, THROWABLE and EXCEPTION types were giving unsupported results specifically and other types were passing the test - [EXAMPLE](https://github.com/ROKUMATE/MetaCall-core/blob/8538bad9e134f54d4a2f775660020d586fa2aa0b/source/tests/metacall_python_conversion_test/source/metacall_python_conversion_test.cpp#L70) you can check here for the testing code if you want and update your code 

also acc. to me i think name test file as `metacall_python_conversion_test` is more good then `metacall_py_conversion_test` as naming convention we following for tests in metacall is specific with `python` and not `py`

viferga:
I started to implement support for type class but it's hard, I wanted to do this for each language for finishing the types left and create a table with all type conversions.

In that way we can document it and also @ROKUMATE we can reason about the ABI breaking change respect to integer types.

viferga:
> also acc. to me i think name test file as `metacall_python_conversion_test` is more good then `metacall_py_conversion_test` as naming convention we following for tests in metacall is specific with `python` and not `py`

Also this is right.

Harsh-dev023:
renamed the test to metacall_python_conversion_test

viferga:
I ran the CI, the ideal would be to progressively implement all types. But if we cannot, it would be good to make it return error or exception when a type is passed.

I did this in node, it makes it silently fail but it's required for tests to pass.

viferga:
@Harsh-dev023 can you merge develop into this branch? The base

viferga:
Sorry I miss-clicked close PR.

The base commit is a broken one. We were merging a big PR. Merge develop into this branch so the CI can run properly.

viferga:
I ran the CI.

Harsh-dev023:
the Python Port tests pass on Ubuntu and Windows, but the macOS Python tests are crashing with a segmentation fault while running python test.py. I'm checking whether it's something introduced by the new type iteration test or an existing issue and will update you once I know more.

viferga:
You need to check here: https://github.com/metacall/core/actions/runs/27414380084/job/81024527223?pr=810

viferga:
It failed because class and object is not implemented properly. We should return exception or throwable instead.

viferga:
I give you an example of what we are looking for: https://github.com/metacall/core/pull/822
At least we should have exceptions for object and class.

Harsh-dev023:
Sir found that Py_IncRef(obj_impl->obj) was being called before the NULL check in TYPE_OBJECT, which was causing the segfault. Moved the check before it, and made both TYPE_CLASS and TYPE_OBJECT return a Python exception instead of NULL when the impl is missing, matching the pattern from #822

PR #814, opened 2026-06-10
Migrate value-returning error sites to metacall_error_throw

ROKUMATE:
## Summary
Replaces `log_write + return NULL` with `metacall_error_throw` across
value-returning entrypoints in `metacall.c` and the loaders, so errors
propagate as throwable values instead of being silently swallowed.

Pattern follows `rapid_json_serial_impl.cpp:422`:

```c
  return metacall_error_throw("metacall", -1, NULL, "<message>", <args>);
```

## Scope
In:
- value-returning functions (return type `void *` / `value`)
- removes the old log + NULL return; single throw statement

Out:
- int-returning functions
- metacall_error_last / metacall_error_clear (still TODO)
- stacktrace propagation
- debug-mode log inside metacall_error_throw

## Progress
- [x] handle-validation entrypoints (metacallhv, metacallhv_s, metacallht_s, metacall_handle_function, metacall_handle_export)
- [ ] "method not implemented" cluster (metacallv_method and friends)
- [ ] remaining value-returning sites in metacall.c
- [ ] per-loader passes (py_loader, node_loader, ...)

### DIFF
--- a/source/loaders/rb_loader/source/rb_loader_impl.c
+++ b/source/loaders/rb_loader/source/rb_loader_impl.c
@@ -13,6 +13,8 @@
 #include <loader/loader.h>
 #include <loader/loader_impl.h>
 
+#include <metacall/metacall_error.h>
+
 #include <portability/portability_path.h>
 
 #include <reflect/reflect_context.h>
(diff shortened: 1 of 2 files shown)


ROKUMATE:
want me to keep adding more conversions to the migration while you 
figure out the rust fmt thing, or hold?

as ig the ci fails are not related to my pr acc. to me

PR #816, opened 2026-06-11
fix(msan): add Google Benchmark to MSan ignorelist

MrSpideyNihal:
Google Benchmark library is fetched via FetchContent and not built with MSan instrumentation, causing false positives in bench tests.

Added `src:*/_deps/googlebenchmark*` to `msan-ignorelist.txt` to suppress false positives from the benchmark library.

Verified locally: set-bench and log-bench pass with this change.

### DIFF
--- a/source/tests/sanitizer/msan-ignorelist.txt
+++ b/source/tests/sanitizer/msan-ignorelist.txt
@@ -13,5 +13,9 @@
 src:*/googletest/*
 fun:testing::*
 
-# Google Benchmark library not built with MSan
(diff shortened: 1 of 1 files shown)


viferga:
https://github.com/metacall/core/commit/c73c808d2443c14490d96937096707384758c40c

viferga:
I am going to ammend it with you as co author

viferga:
https://github.com/metacall/core/commit/1c9f98b8ae84902c3a16ced6c83bb6f2f9782a36

PR #822, opened 2026-06-14
tests: rb: Add type conversion test for ruby loader

AshutoshSharma-pixel:
Adds metacall_ruby_conversion_test, modeled after metacall_node_conversion_test.

The test loads an identity function from memory using the rb loader and runs it
against all METACALL_* types, printing the conversion matrix.

Local test output on macOS with Ruby 3.2.2:

Boolean  => Boolean
Fixnum   => Fixnum
Bignum   => Bignum
Float    => Float
String   => String
Array    => Array
NilClass => NilClass

Notable difference from the Node loader: Ruby preserves numeric type granularity
(Fixnum, Bignum, Float remain distinct) whereas Node collapses all numeric types
to METACALL_DOUBLE.

Tested locally with:
cmake -DOPTION_BUILD_LOADERS=ON -DOPTION_BUILD_LOADERS_RB=ON -DOPTION_BUILD_TESTS=ON
cmake --build . --target metacall-ruby-conversion-test
ctest -R metacall-ruby-conversion-test --verbose

[ PASSED ] 1 test


viferga:
Running CI

viferga:
Add tests please.

AshutoshSharma-pixel:
added assertions for Bool, Int, Long, Double, String, Array, Map and Buffer round-trips. also implemented Buffer (IO::Buffer), Array and Map serialization/deserialization in rb_loader. all pass locally. [ PASSED ] 1 test

AshutoshSharma-pixel:
fixed the ubuntu:jammy build failure ruby/io/buffer.h is only available in Ruby >= 3.0, so I've gated the IO::Buffer include and usage behind RUBY_API_VERSION_MAJOR >= 3. on older Ruby versions the buffer falls back to a plain binary string. also added ruby/version.h to the loader include header to ensure the macro is defined. still passes locally.

viferga:
Cool, running the CI.

AshutoshSharma-pixel:
fixed the formatting check failure clang-format 12 expects \t#include inside #if blocks, not #\tinclude. one character fix.

AshutoshSharma-pixel:
The Ubuntu Jammy failed again i see this time i tried this 
the actual issue: RUBY_API_VERSION_MAJOR >= 3 only checks the Ruby version macro, not whether ruby/io/buffer.h is actually installed some distro Ruby packages ship the 3.x runtime without that specific dev header. switched to CMake's check_include_files to detect real header availability (HAVE_RUBY_IO_BUFFER), combined with the existing version check for extra safety, and falls back to a plain binary string when the header is unavailable. tested locally on macOS, it still passes the full conversion matrix.

AshutoshSharma-pixel:
reverted the CMake check and switched to #if RUBY_API_VERSION_MAJOR >= 3 && RUBY_API_VERSION_MINOR >= 1  IO::Buffer was introduced in Ruby 3.1.0 so the minor version check is the right guard here. cleaner and more reliable than header detection.

Issue #123, opened 2021-04-22
Make C# Loader production ready


viferga:
## 🚀 Feature
Right now the C# Loader does not follow the interface of MetaCall properly. This is very problematic because all scripts are loaded into the current `AppDomain` and this will allow collisions between dependencies. Apart from that, there is another problem related to the dependencies itself, they are usually not tracked properly and fail to load because the AppDomain cannot locate them. I have been playing around with them with a PoC of Blazor, and it shows an error related to strong naming (https://stackoverflow.com/questions/16674315/understanding-appdomains-and-strong-naming).

Here's an explanation of how to load assemblies with dependencies: https://stackoverflow.com/a/37060921
The repository of the PoC is here (I have attached it as a zip too)
[ResolveDependencies-master.zip](https://github.com/metacall/core/files/6357832/ResolveDependencies-master.zip)
: https://github.com/mathume/ResolveDependencies

The current implementation does not return a handle when loading, instead all functions are loaded into a singleton. As you can see, it returns boolean instead of a handle pointer.
 1) C# Side: https://github.com/metacall/core/blob/4baa35658e316ba8848a2bf966ec09197a7bfef6/source/loaders/cs_loader/netcore/source/MetacallEntryPoint.cs#L93
 2) NetCore Interface: https://github.com/metacall/core/blob/4baa35658e316ba8848a2bf966ec09197a7bfef6/source/loaders/cs_loader/source/netcore.cpp#L143
 3) Loader Interface: https://github.com/metacall/core/blob/4baa35658e316ba8848a2bf966ec09197a7bfef6/source/loaders/cs_loader/source/cs_loader_impl.c#L269

This breaks the design, it should return a structure in the 3) containing a pointer to the handle returned by 1), which should contain a pointer to the new AppDomain (one by handle) and a pointer to the list of loaded assemblies or similar, so later on the functions can be retrieved and they are scoped by handle, instead of by loader instance like now: https://github.com/metacall/core/blob/4baa35658e316ba8848a2bf966ec09197a7bfef6/source/loaders/cs_loader/netcore/source/MetacallEntryPoint.cs#L76

The good part is that the function container already points to their own assembly so it will be simpler to migrate: https://github.com/metacall/core/blob/4baa35658e316ba8848a2bf966ec09197a7bfef6/source/loaders/cs_loader/netcore/source/FunctionContainer.cs#L18

[...]

PR #825, opened 2026-06-20
cs_loader:  #123 follow-ups  net7.0 tests + CTest wiring + cleanup 

G00dS0ul:
# Description

Addresses the follow-up checklist @viferga posted on #123 ([comment](https://github.com/metacall/core/issues/123#issuecomment-4661930445)). Four of the five items are complete; task 5 is **scaffolded but gated** because it surfaced a pre-existing loader bug (details below).

## Type of change

# Task Status

- [x] Delete `LoaderV2.contextWeakRef`
- [x] Move unit tests into `source/Test`
- [x] Downgrade test project net10.0 → net7.0
- [x] Wire `dotnet test` into CMake (CTest) 
- [ ] Function-call integration test in `tools/docker/tests`

## What's done

**1. Removed dead `contextWeakRef`** — after the per-script ALC rewrite, `UnloadScript` uses a local `WeakReference`, so the field was never read. Removed from `Providers/LoaderV2.cs`.

**2. Moved tests** → `source/loaders/cs_loader/netcore/source/Test/`, with the `<ProjectReference>` path to `project.csproj` fixed for the new depth. Keeps the test project inside the `source/` tree with the rest of the managed code.

**3. Retargeted tests to net7.0** to match the runtime MetaCall ships. Pinned net7.0-compatible packages: `Microsoft.NET.Test.Sdk` 17.6.3, `xunit` 2.4.2, `xunit.runner.visualstudio` 2.4.5, `coverlet.collector` 3.2.0. Also added `<Compile Remove="Test\**\*.cs" />` to the main project so its default glob doesn't pull the test files in.

**4. CTest wiring** — registered `cs_loader_test` in `netcore/CMakeLists.txt` (gated behind `OPTION_BUILD_TESTS`), following the rs_port pattern, with `DOTNET_ROLL_FORWARD=Major` so the net7.0 testhost runs on a newer runtime if 7.0 isn't installed.

**Bonus:** small `cmake/FindDotNET.cmake` robustness fix  added the missing `include(FindPackageHandleStandardArgs)` so a trimmed-down configure doesn't fail with *"unknown command `find_package_handle_standard_args`"*. 

## What's NOT done — task 5 (gated)

I added `Sum.cs` and a C# block in `tools/docker/tests/scripts/index.py`/`index.js`, but **left the block commented out** (mirroring the existing Java block). Reason: building against `metacall/core:cli` shows the **cs_loader crashes on init** (`std::terminate`, exit 134) *before any script compiles*, so C# is currently non-functional in the packaged cli image.

The scaffold is in place so the test can be flipped on the moment the loader is fixed.

# Checklist:

- [x] I have tested the tests implicated (if any) by my own code and they pass (`ctest -R cs_loader_test --output-on-failure`).

## Testing

- All 9 xUnit tests pass on net7.0 locally (`dotnet test`).
- `ctest -R cs_loader_test` passes locally and in WSL.
- `docker build` of `tools/docker/tests` passes with the C# block gated.


.gitignore
```diff
@@ -64,3 +64,12 @@ _start-cmd.bat
 
 # Auto-generated bake file
 docker-compose.bake.yml
+
+.venv/
```
viferga: Respect to this.. isn't the .gitignore here covering this already?

https://github.com/metacall/core/blob/develop/source/loaders/cs_loader/netcore/source/.gitignore

viferga: The rest is perfect, good job!

PR #839, opened 2026-07-02
feat(ruby): enable Clang ASan/TSan Ruby instrumentation

MrSpideyNihal:
## Summary

This PR enables Ruby to be built from source with Clang when running AddressSanitizer (ASan) and ThreadSanitizer (TSan) builds.

Previously, the sanitizer jobs installed the system Ruby package, which is not instrumented. As a result, the Ruby loader could not be properly tested under Clang sanitizers.

When Clang + ASan/TSan is requested, Ruby is now compiled from source using `rbenv` and `ruby-build`, allowing it to be instrumented with the same sanitizer flags used by MetaCall.

## Changes

- add `test-clang-address-sanitizer`
- add `test-clang-thread-sanitizer`
- build Ruby 3.4.1 from source when:
  - `INSTALL_CLANG=1`
  - `INSTALL_ADDRESS_SANITIZER=1` or `INSTALL_THREAD_SANITIZER=1`
- use the system `clang` compiler instead of hardcoding `clang-18`
- configure Ruby with:
  - `USE_MN_THREADS=0`
  - `--disable-install-doc`
- configure sanitizer-specific `CFLAGS` and `LDFLAGS`
- print ruby-build logs if Ruby compilation fails
- keep the existing package installation path unchanged for non-Clang builds

## Testing

Tested locally by running:

```bash
./docker-compose.sh test-clang-address-sanitizer


viferga:
I have implemented the instrumentation properly, have a look at the comments and doc I left in the environment.sh:
https://github.com/metacall/core/commit/29f9e4d93c89daf351c1006d02fe2c969ffa52de

PR #845, opened 2026-07-13
Adding Retry and Timeout for rpc remote discovery

Mag-D-Anas:


viferga:
@Mag-D-Anas it has conflicts.

viferga:
Also it's not tested, we could try a test maybe.. with a special service that fails to connect the first 3 times.

viferga:
I have to discard this PR in favor of this commit: https://github.com/metacall/core/commit/061064c77c958b95ab34934b09a36e6f4fdfc818

The environment variables is not what I wanted, the idea is that the configuration now includes the intervals...

I have added tests for the timeout and retry, please check them out.

PR #846, opened 2026-07-16
Mingw/msys2 support

yug105:
Adds MinGW/MSYS2 support for Windows. A new CI workflow builds and tests on `mingw64`, `mingw32`, `ucrt64`, and `clang64` using the `tools/metacall-*.sh` scripts. All four environments pass 100% of the tests.

Changes:

* `syslog`: Fixed the `ReportEvent` const pointer type.
* `metacall_link`: Added `WINAPI` to the trampoline typedef to fix a calling-convention mismatch on 32-bit Windows.
* Fork test: Guarded `pid_t` with `_MSC_VER`, as MinGW already defines it.
* Linked `dbghelp` for `plthook_detour` and skipped `-rdynamic` on Windows.
* Built GTest statically on Windows to avoid DLL-not-found failures.
* Serialization test: Added pointer-format cases for each MSYS2 runtime.

Disabled on MinGW because they are not currently supported:

* **Fork safety:** Not supported on Windows; MSVC already disables it as well. Detour fork support requires the delayed-load mechanism.
* **Backtrace plugin:** `backward-cpp` fails to build with GCC and deadlocks during loading with Clang.


viferga:
@yug105 respect to this code:

```
if(MINGW)
	set(OPTION_BUILD_PLUGINS_BACKTRACE OFF)
endif()
```

I have removed it, you must go to the backtrace plugin, I am going to add that code but in theory it seems that it supports the backtrace one:
https://github.com/bombela/backward-cpp/blob/0bfd0a07a61551413ccd2ab9a9099af3bad40681/BackwardConfig.cmake#L206

What was the error you were getting?

Edit: Here is how to do it:
https://github.com/yug105/core/blob/c155e69e82df6138541ccba4ae08e9fd8a5efef3/source/plugins/backtrace_plugin/CMakeLists.txt#L6

viferga:
I also have removed this:

```
# Fork safety is not supported on Windows yet (detour requires the delayed load mechanism)
if(WIN32)
	set(FORK_SAFE_DEFAULT OFF)
else()
	set(FORK_SAFE_DEFAULT ON)
endif()
option(OPTION_FORK_SAFE "Enable fork safety." ${FORK_SAFE_DEFAULT})
```

It's not the proper way of doing it. What error were you getting with this?

Respect to the delay load mechanism, it's documented on metacall/plthook-poc, I can write the function to reuse it automatically for you, so we can add this to all cases where we need the delay load mechanism.

But the delay load mechanism is only required in loaders, the fork should not need it. Fork safety mechanism works on MSVC.

tools/metacall-build.sh
```diff
@@ -84,7 +84,7 @@ sub_build() {
 	else
 		# Tests (coverage needs to run the tests)
 		if [ $BUILD_TESTS = 1 ] || [ $BUILD_BENCHMARKS = 1 ] || [ $BUILD_COVERAGE = 1 ]; then
-			ctest -j$(getconf _NPROCESSORS_ONLN) --timeout 5400 --output-on-failure --test-output-size-failed 3221000000 -C $BUILD_TYPE
```
viferga: Is there any reason for this change?

yug105: Yes, actually this error was coming `Invalid value for '--test-output-size-failed': 3221000000` so i changed it to 2147483647 (INT_MAX) to make it valid on windows

yug105:
> I also have removed this:
> 
> ```
> # Fork safety is not supported on Windows yet (detour requires the delayed load mechanism)
> if(WIN32)
> 	set(FORK_SAFE_DEFAULT OFF)
> else()
> 	set(FORK_SAFE_DEFAULT ON)
> endif()
> option(OPTION_FORK_SAFE "Enable fork safety." ${FORK_SAFE_DEFAULT})
> ```
> 
> It's not the proper way of doing it. What error were you getting with this?
> 
> Respect to the delay load mechanism, it's documented on metacall/plthook-poc, I can write the function to reuse it automatically for you, so we can add this to all cases where we need the delay load mechanism.
> 
> But the delay load mechanism is only required in loaders, the fork should not need it. Fork safety mechanism works on MSVC.

The error occurred on MinGW:

`PLTHook failed to open (code 6, path ntdll.dll): ImageDirectoryEntryToData error`

I temporarily disabled `fork` behind an `if(WIN32)` guard to get the MinGW build passing. However, that condition also applies to MSVC, where disabling `fork` was incorrect, the underlying MinGW issue still needs a proper fix.

yug105:
> @yug105 respect to this code:
> 
> ```
> if(MINGW)
> 	set(OPTION_BUILD_PLUGINS_BACKTRACE OFF)
> endif()
> ```
> 
> I have removed it, you must go to the backtrace plugin, I am going to add that code but in theory it seems that it supports the backtrace one: https://github.com/bombela/backward-cpp/blob/0bfd0a07a61551413ccd2ab9a9099af3bad40681/BackwardConfig.cmake#L206
> 
> What was the error you were getting?
> 
> Edit: Here is how to do it: https://github.com/yug105/core/blob/c155e69e82df6138541ccba4ae08e9fd8a5efef3/source/plugins/backtrace_plugin/CMakeLists.txt#L6

```
FAILED: plugins/backtrace_plugin/libbacktrace_plugind.dll

ld.exe: backtrace_plugin.cpp.obj: undefined reference to `__imp__set_abort_behavior`

ld.exe: libbackward.a(backward.cpp.obj): SignalHandling constructor: undefined reference to `__imp__set_abort_behavior`

collect2.exe: error: ld returned 1 exit status
```
this is the error i got, thanks for the link i think it will get away with the link error we are getting.

source/plugins/backtrace_plugin/CMakeLists.txt
```diff
@@ -3,6 +3,10 @@ if(NOT OPTION_BUILD_LOADERS OR NOT OPTION_BUILD_LOADERS_EXT OR NOT OPTION_BUILD_
 	return()
 endif()
 
+if(MINGW)
+	add_compile_definitions(__MSVCRT_VERSION__=0x900)
```
viferga: you can make this per target instead of per project

PR #851, opened 2026-07-25
Initial rust template generics and template test

SATVIKsynopsis:
# Description

Added the initial Rust generics template implementation and tests.

## Type of change

- [ ] Documentation update

# Checklist:


source/loaders/rs_loader/rust/compiler/src/lib.rs
```diff
@@ -369,6 +369,8 @@ pub struct Function {
     name: String,
     ret: Option<FunctionParameter>,
     args: Vec<FunctionParameter>,
+    #[allow(dead_code)]
```
viferga, line 373: why dead code?

viferga (commented):
It's a good kickstart but we have to think yet how to handle templates over types. Either we can use the clang style, like FunctionTemplate and ClassTemplate types, or create a generic type called Template.. or make Function and Class have template types by default.. I am not sure how to handle it in the best way.

We will need a new interface for instantiate the templates too. So we will need some kind of codegen on the loader side.

source/loaders/rs_loader/rust/compiler/src/lib.rs
SATVIKsynopsis: i added it temporarily because generics is only used for the template handling flow right now, will remove it once the template representation is fully integrated.

PR #862, opened 2026-09-01
Fix memory leaks and Generalize map key parsing in Go port

roshankumar0036singh:
# Pull Request: Fix Memory Leaks in Go Port

## Overview
This PR addresses several memory leaks related to `metacall_value` allocations in the Go Port, specifically during map creation and test assertions.

*Note on map keys*: A regression test (`TestMapNonStringKeys`) has been added to demonstrate a panic that occurs when the underlying engine (like Python) returns a dictionary with integer keys. 

### 1. Fix memory leak during Map conversion (`source/ports/go_port/source/go_port.go`)
Added `defer C.free(unsafe.Pointer(cArgs))` to prevent leaking the C array of tuples when converting a `reflect.Map`.

### 2. Expose `valueDestroy` utility (`source/ports/go_port/source/go_port.go`)
Exposed a lightweight wrapper for cleaning up `metacall_value` pointers. Added a `nil` pointer guard (`if v == nil { return }`) so this wrapper can be safely called defensively in `defer` statements or after error paths where the pointer might not have been fully allocated.

### 3. Prevent test values memory leak (`source/ports/go_port/source/go_port_test.go`)
Added the `valueDestroy(ptr)` call at the end of the `TestValues` loop to clean up the C values allocated by `goToValue(tt.input, &ptr)`.

### 4. Added regression test for non-string map keys (`source/ports/go_port/source/go_port_test.go`)
Added `TestMapNonStringKeys` which currently panics with `interface conversion: interface {} is int, not string`. This test is included to highlight the issue for future resolution.

### DIFF
--- a/source/ports/go_port/source/go_port.go
+++ b/source/ports/go_port/source/go_port.go
@@ -603,18 +603,19 @@ func valueToGo(value unsafe.Pointer) interface{} {
 				pairs[i] = KeyValuePair{Key: key, Value: val}
 
 				keyType := reflect.TypeOf(key)
+				if keyType != nil && !keyType.Comparable() {
+					panic(fmt.Errorf("metacall: map key type %v is not usable as a Go map key", keyType))
+				}
+
 				if i == 0 {
 					uniformType = keyType
-					if uniformType == nil || !uniformType.Comparable() {
-						isUniform = false
-					}
 				} else if isUniform && keyType != uniformType {
 					isUniform = false
 				}
 			}
 
 			// Second pass: construct the map
-			if isUniform && uniformType.Kind() == reflect.String {
+			if isUniform && uniformType != nil && uniformType.Kind() == reflect.String {
 				// Fast path: map[string]interface{}
 				m := make(map[string]interface{}, size)
 				for _, pair := range pairs {
@@ -622,6 +623,7 @@ func valueToGo(value unsafe.Pointer) interface{} {
 				}
 				return m
 			} else if isUniform {
+				// TODO: revisit with Go generics
 				// Reflect uniform path: map[T]interface{}
 				mapType := reflect.MapOf(uniformType, reflect.TypeOf((*interface{})(nil)).Elem())
 				m := reflect.MakeMapWithSize(mapType, int(size))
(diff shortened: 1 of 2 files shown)


source/ports/go_port/source/go_port.go
```diff
@@ -636,3 +636,10 @@ func Destroy() {
 	// Wait for all work to complete
 	wg.Wait()
 }
+
+func valueDestroy(v unsafe.Pointer) {
+	if v == nil {
```
viferga: unless nil in Go is different from 0x0, then this is redundant

viferga: you can directly use: C.metacall_value_destroy(v)

source/ports/go_port/source/go_port_test.go
```diff
@@ -244,3 +246,25 @@ func BenchmarkNodeJSParallel(b *testing.B) {
 		}
 	})
 }
+
+func TestMapNonStringKeys(t *testing.T) {
+	input := map[int]string{
+		1: "one",
```
viferga: if this is valid in Go, we can generalize it

roshankumar0036singh: The fix i am  planning (fmt.Sprintf("%v", key)) is generic, so it'll handle any key type, not just int

viferga: you're going to convert any type to string then?

viferga: I mean any key type

roshankumar0036singh: there  are two approaches i could think of that 
1> Convert any non-string key with fmt.Sprintf("%v", key), keep returning map[string]interface{} it will not have any breaking changes but different key types can collide into the same string     

  2> Return an error (or change the function signature) instead of coercing, when a map has non-string keys  but valueToGo doesn't currently return an error so this  means changing its signature (breaking for callers) or panicking with a clearer message

roshankumar0036singh: or we could  inspect the key's reflect.Kind()  construct map[string]interface{} when keys are strings, falling back to map[interface{}]interface{} (or some other generic form) only when they aren't That preserves the existing contract for the common path

source/ports/go_port/source/go_port.go
roshankumar0036singh: updated it

source/ports/go_port/source/go_port_test.go
roshankumar0036singh: @viferga can you review the current two pass plan although its time consumer but as per your request of detecting the kesy at run time so   one type => one type, else interface => interfacee does it holds well

viferga: it looks good, let me run the ci

roshankumar0036singh:
@viferga hi is everything alright in the ci ?

viferga:
Sorry I forgot to merge.

PR #875, opened 2026-09-09
Go port fix

Schrodinger257:
# Description

This PR aim is to fix go port and make it usable for production.
currently, i fixed memory leak with map creation and goCallbacksPtr not removed when metacallfv_await_struct_s fail.
also fixed go routine leaks when shutting down and unified the works with safeWork interface and add context for workers to be able to drain queue before shutdown and made Destroy() wait for DestroyUnsafe() that calls metacall_destroy() until it finishes for a safe shutdown .
Current Known Goals:
- [x] consider using cgo.handle instead of pointer.go file
- [x] add support of address and thread sanitizers (managed by go runtime, added profiles and trace support)
- [x] add support of more types, generic Map is already implemented in PR #862  (uint32, uint64, uintptr, ...)
- [x] add support of advanced metacall types (class, object, ...)
- [x] try to add support of remaining metacall APIs if they are needed (inspect, clear, logs, load from package, ...)
- [x] update README of go port to reflect new changes

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist:

- [x] I have tested my code with `OPTION_BUILD_ADDRESS_SANITIZER` or `./docker-compose.sh test-address-sanitizer &> output` and `OPTION_TEST_MEMORYCHECK`.
- [x] I have tested my code with `OPTION_BUILD_THREAD_SANITIZER` or `./docker-compose.sh test-thread-sanitizer &> output`.


viferga:
Hello, can you check there's no conflicts with https://github.com/metacall/core/pull/862 ?

Schrodinger257:
> Hello, can you check there's no conflicts with #862 ?

No there's no conflicts

viferga:
Can you maintain pointer.go as it was, and add class.go and object.go separately?

Schrodinger257:
> Can you maintain pointer.go as it was, and add class.go and object.go separately?

okay, i will separate class and object and future. the pointer.go file do you want to reuse it instead of cgo.handle ? is there a reason ? cgo.handle should be better as it guarantees avoiding breaking pointer-passing rules of cgo

viferga:
What's cgo.handle?

Schrodinger257:
> What's cgo.handle?

It is a type that allow you to pass pointers or values containing pointers like string between go and c without violating pointer-passing rules and it make sure that go garbage collector won't free or reclaim it until you call handle.Delete(). It converts the value given to it (the pointer) into an uintptr then pass it to c and after receiving it, it converts it again to the original value (pointer) which introduce a problem which is c can't modify or dereference the handle as it is an uintptr not a pointer. The pointer.go file was doing the same work as cgo.handle but i thought using a built-in type from go is better than using a custom one.

viferga:
> > What's cgo.handle?
> 
> It is a type that allow you to pass pointers or values containing pointers like string between go and c without violating pointer-passing rules and it make sure that go garbage collector won't free or reclaim it until you call handle.Delete(). It converts the value given to it (the pointer) into an uintptr then pass it to c and after receiving it, it converts it again to the original value (pointer) which introduce a problem which is c can't modify or dereference the handle as it is an uintptr not a pointer. The pointer.go file was doing the same work as cgo.handle but i thought using a built-in type from go is better than using a custom one.

If it works, don't touch it.

viferga:
I just checked and it was published this month:
https://pkg.go.dev/runtime/cgo@master#Handle

I think it may be too new, forcing consumers to update to latest go version, I prefer if we maintain support to older versions until this is more widespread.

Another thing is, in the future I want to completely get rid of cgo. It's really slow and has other problems attached to it.

In other software I'm using this as alternative:
https://github.com/ebitengine/purego

Schrodinger257:
> I just checked and it was published this month: https://pkg.go.dev/runtime/cgo@master#Handle
> 
> I think it may be too new, forcing consumers to update to latest go version, I prefer if we maintain support to older versions until this is more widespread.
> 
> Another thing is, in the future I want to completely get rid of cgo. It's really slow and has other problems attached to it.
> 
> In other software I'm using this as alternative: https://github.com/ebitengine/purego

i understood your point of view. i will return to use pointer.go. as for purego instead of cgo, do you suggest using it for the whole go port ?

viferga:
> > I just checked and it was published this month: https://pkg.go.dev/runtime/cgo@master#Handle
> > I think it may be too new, forcing consumers to update to latest go version, I prefer if we maintain support to older versions until this is more widespread.
> > Another thing is, in the future I want to completely get rid of cgo. It's really slow and has other problems attached to it.
> > In other software I'm using this as alternative: https://github.com/ebitengine/purego
> 
> i understood your point of view. i will return to use pointer.go. as for purego instead of cgo, do you suggest using it for the whole go port ?

We will address that in another moment, optimizations and things like that go to the end, and before optimizing anything we need to setup profiling.

Schrodinger257:
> > > I just checked and it was published this month: https://pkg.go.dev/runtime/cgo@master#Handle
> > > I think it may be too new, forcing consumers to update to latest go version, I prefer if we maintain support to older versions until this is more widespread.
> > > Another thing is, in the future I want to completely get rid of cgo. It's really slow and has other problems attached to it.
> > > In other software I'm using this as alternative: https://github.com/ebitengine/purego
> > 
> > 
> > i understood your point of view. i will return to use pointer.go. as for purego instead of cgo, do you suggest using it for the whole go port ?
> 
> We will address that in another moment, optimizations and things like that go to the end, and before optimizing anything we need to setup profiling.

so for now should i use purego in future only ?

viferga:
> > > > I just checked and it was published this month: https://pkg.go.dev/runtime/cgo@master#Handle
> > > > I think it may be too new, forcing consumers to update to latest go version, I prefer if we maintain support to older versions until this is more widespread.
> > > > Another thing is, in the future I want to completely get rid of cgo. It's really slow and has other problems attached to it.
> > > > In other software I'm using this as alternative: https://github.com/ebitengine/purego
> > > 
> > > 
> > > i understood your point of view. i will return to use pointer.go. as for purego instead of cgo, do you suggest using it for the whole go port ?
> > 
> > 
> > We will address that in another moment, optimizations and things like that go to the end, and before optimizing anything we need to setup profiling.
> 
> so for now should i use purego in future only ?

No, don't use purego at all.

viferga:
Sorry I miss-clicked.

Restore pointer.go, add the new features in class.go, object.go and future.go.

Add tests and integrate those tests with cmake. But also try to make small PRs, right now it doesn't matter because most of the things you're adding is new code, but if you change existing code like in the bugs you reported before, it's better to do it in small parts so it's better to review.

Schrodinger257:
> Sorry I miss-clicked.
> 
> Restore pointer.go, add the new features in class.go, object.go and future.go.
> 
> Add tests and integrate those tests with cmake. But also try to make small PRs, right now it doesn't matter because most of the things you're adding is new code, but if you change existing code like in the bugs you reported before, it's better to do it in small parts so it's better to review.

integration of cmake tests is already done.

source/ports/go_port/source/class.go
```diff
@@ -0,0 +1,128 @@
...
+	})
+	return cls
+}
+
+func (c *Class) New(name string, args ...interface{}) (*Object, error) {
+	if c.ptr == nil {
+		return nil, errors.New("can't use nil ptr for class creation")
+	}
+
+	defer runtime.KeepAlive(c)
```
viferga, line 44: why do you do this? this will leak the class no?

Schrodinger257: no this tells garbage collector to keep it alive during execution of StaticSet() or StaticGet() to prevent the problem in this PR #888 
all objects, classes and futures finalizers are called when calling Destroy() and they are freed or when they are no longer referenced. so no leaks

Issue #896, opened 2026-09-21
adt: vector_insert_empty shifts elements in the wrong direction (memmove src/dst swapped)


bhuvan-somisetty:
## 🐛 Bug Report

`vector_insert_empty` (source/adt/source/adt_vector.c) has the `memmove` source and destination swapped compared to `vector_insert`, so it does not open a gap at `position`.

### Expected Behavior
`vector_insert_empty(v, position)` with `position < size` should move elements `[position, size)` one slot to the right (like `vector_insert`), leaving an empty slot at `position`.

### Current Behavior
```c
memmove(vector_data_offset_bytes(v, position * v->type_size),        // dst
        vector_data_offset_bytes(v, (position + 1) * v->type_size),  // src
        (v->size - position) * v->type_size);
```
This copies elements after `position` one slot to the left, overwriting the element at `position`, and the length `(size - position)` reads one element past the last valid one. `vector_insert` does it the correct way round (dst = position + 1, src = position).

### Possible Solution
Swap the arguments, as in `vector_insert`:
```c
memmove(vector_data_offset_bytes(v, (position + 1) * v->type_size),
        vector_data_offset_bytes(v, position * v->type_size),
        (v->size - position) * v->type_size);
```
A small test in the adt test suite (insert in the middle and at the front, then check order) would cover it.

### Steps to Reproduce
Found by code inspection, not yet reproduced at runtime.
1. Create a vector of `int` with elements `10, 20, 30`.
2. Call `vector_insert_empty(v, 0)`.
3. Expected: `_, 10, 20, 30`. Actual (from the memmove arguments): the element at index 0 is overwritten by `20`, giving `20, 30, 30` instead.

### Context (Environment)
- Branch `develop`, commit 4cad3873.
- Current callers (`file_loader_impl.c:261`, `:430`) only insert into an empty vector or at the end, so it is latent today, but the API is public (`adt_vector.h:530`).
- Happy to send a fix with a test if a maintainer agrees on the scope.

bhuvan-somisetty:
Hi @viferga, I spotted this while reading through the adt vector code and it looks like a small, self-contained fix. I'd like to work on it (swap the memmove args and add a test that covers front/middle insertion). Would that be okay with you, and is that scope fine? I'll wait for your go-ahead before opening a PR. Thanks!

viferga:
[...]

PR #897, opened 2026-09-21
adt: add test for vector_insert_empty (refs #896)

bhuvan-somisetty:
Refs #896. As discussed there, this first adds a unit test in `adt_vector_test` that checks `vector_insert_empty` opens a gap at front and middle positions and keeps the other elements in order. It should fail on CI with the current code; I'll push the memmove fix once it does.


source/tests/adt_vector_test/source/adt_vector_test.cpp
```diff
@@ -52,3 +52,36 @@ TEST_F(adt_vector_test, DefaultConstructor)
...
+
+		ASSERT_EQ((size_t)vector_size(v), (size_t)(count + 1));
+
+		size_t marker = 100;
+
+		memcpy(vector_at(v, position), &marker, sizeof(size_t));
+
+		for (size_t i = 0; i < count + 1; ++i)
+		{
+			size_t expected = (i < position) ? i : ((i == position) ? marker : i - 1);
```
viferga: make here the for until count and check count+1 outside of the for please, it will make things more readable

bhuvan-somisetty: Done, the loop now runs up to count and the last shifted element is checked right after it. Thanks!

viferga: but now change the condition of expected, no?

bhuvan-somisetty: Yep, you're right, the nested condition was hiding that. I split it into before the gap, the marker, and after the gap, so there's no expected ternary anymore. Pushed.

viferga:
I still feel the test is a bit confusing. I would do 3 tests:

Test 1:
```
0 1 2 3 4 5 6 7 8 9
insert at 0

verify that we have:
0 0 1 2 3 4 5 6 7 8 9
```

Test 2:
```
0 1 2 3 4 5 6 7 8 9

insert at 5
0 1 2 3 4 5 5 6 7 8 9
```

Test 3:
```
insert at the end:
0 1 2 3 4 5 6 7 8 9 [here we will have garbage]

verify first 10 are correct
```

bhuvan-somisetty:
Yeah, fair, that loop was harder to read than it needed to be. I split it into three tests like you laid out: front (insert at 0), middle (insert at 5) and end (insert at 10, only checking the first 10 since the new slot is garbage). Each one starts from 0..9 and just checks the result directly. Pushed. I haven't built it locally, so I'm counting on CI, and once it goes red I'll push the memmove fix.

bhuvan-somisetty:
Pushed the fix too: the memmove had source and destination swapped, so insert_empty was shifting things the wrong way. The size argument was already right. The three tests should cover it now.

viferga:
@bhuvan-somisetty Can you contact me via discord?

bhuvan-somisetty:
Sure!

